PR #4, opened 2024-10-16
feat: short answer reset button

ArturGaspar:
## Description

Adds a reset button to the short answer XBlock.

image

## Testing instructions

1. Enable the option
2. Use the XBlock as a learner
3. Test that you are able to reset the chat
4. Disable the option
5. Use the XBlock as a learner
6. See that the reset button is gone

## Other information

Private-ref: https://tasks.opencraft.com/browse/BB-9298

### DIFF
--- a/ai_eval/templates/shortanswer.html
+++ b/ai_eval/templates/shortanswer.html
@@ -19,6 +19,9 @@
             </div>
         </div>
         <div class="submit-row">
+            {% if self.allow_reset %}
+                <span id="reset-button" class="disabled-btn">Reset chat</span>
+            {% endif %}
             <textarea class="user-input" rows="1" placeholder="Type your answer here" maxlength="1000"></textarea>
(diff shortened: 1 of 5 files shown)


ai_eval/templates/shortanswer.html
```diff
@@ -19,6 +19,9 @@
             </div>
         </div>
         <div class="submit-row">
+            {% if self.allow_reset %}
+                <span id="reset-button">Reset</span>
```
pkulkark, line 23: Nit: I think we should name this `Reset Chat` or `Clear chat` instead of just `Reset` so the operation is clearer.

pkulkark:
@ArturGaspar Would it be possible to activate the button only when there is an answer submitted (or chat exists). In other words, after the first submit?

pkulkark (approved):
LGTM :+1:

- [x] I tested this: verified the reset button works as expected.
- [x] I read through the code
- ~[ ] I checked for accessibility issues~
- ~[ ] Includes documentation~

PR #5, opened 2024-10-16
feat: character image in short answer XBlock

ArturGaspar:
## Description

Adds an optional, configurable image to the left of the short answer XBlock, representing a character the learner is dialoguing with.

image

## Testing instructions

1. Set "Character Image URL" in the XBlock settings
2. See that it works

## Other information

Private-ref: https://tasks.opencraft.com/browse/BB-9245

### DIFF
--- a/ai_eval/shortanswer.py
+++ b/ai_eval/shortanswer.py
@@ -30,6 +30,15 @@ class ShortAnswerAIEvalXBlock(AIEvalXBlock):
         scope=Scope.settings,
     )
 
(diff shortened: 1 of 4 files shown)


pkulkark (approved):
LGTM :+1: 

- [x] I tested this: Verified that the character image shows up in the xblock when set.
- [x] I read through the code
- [ ] ~I checked for accessibility issues~
- [ ] ~Includes documentation~

PR #8, opened 2025-02-26
 feat: allow specifying API keys globally for site and instance 

Agrendalath:
## Description

This implements the option to specify the API key for each model via:
1. Django admin (Site Configurations).
2. Django settings.

## Testing instructions

1. If you are using Tutor, you can create the following YAML [plugin](https://docs.tutor.edly.io/plugins/intro.html) to test this:
    ```yml
    name: dev_config
    version: 0.1.0
    patches:
      openedx-common-settings: |
        SITE_CACHE_TTL = 0  # To avoid cache delays after updating Site Configuration.
        XBLOCK_SETTINGS = {
           "ai_eval": {
               "GPT4O_API_KEY": "your-openai-api-key",
            }
        }
    ````
1. Use the instructions from the README to verify that the functionality works. If anything is missing, we should update them.

## Other information

In the first two commits, I:
1. Dropped the Python 3.8 support, because it didn't support `Self`. I double-checked that the client does not use this version.
3. Converted existing tests to `pytest` to avoid mixing `mark.parametrize` from `test_compat.py` and `ddt` in the same repository. It's a general platform's direction to use `pytest` tests.

_Private-ref: [BB-9551](https://tasks.opencraft.com/browse/BB-9551)_


README.md
```diff
@@ -32,7 +32,36 @@ This repository hosts two Open edX XBlocks:
...
+       "ai_eval": {
+           "GPT4O_API_KEY": "your-openai-api-key",
+        }
+    }
+    ```
+
+#### Security Considerations
+
+For better security, we recommend using site configuration or Django settings instead of configuring API keys at the 
+XBlock level. This prevents API keys from being exposed in course exports.
```
ArturGaspar, line 62: @Agrendalath  Thank you for noting this!

Agrendalath:
@ArturGaspar, sure thing - added in daf785b.

ArturGaspar (approved):

PR #9, opened 2025-02-28
feat: multi-agent AI XBlock

ArturGaspar:
## Description

Adds an XBlock similar to the short answer one, but which can select between different instructions for the LLM depending on context.

## Testing instructions

1. Add "multiagent_ai_eval" and configure required settings
2. Use the XBlock so that it will switch characters.
3. Use the XBlock so that it will end the session and provide an evaluation report.

The below example with the default settings shows how to test this functionality.

image

## Other information

Private-ref: https://tasks.opencraft.com/browse/BB-9325

### DIFF
--- a/ai_eval/multiagent.py
+++ b/ai_eval/multiagent.py
@@ -1,0 +1,636 @@
+"""Multi-agent AI XBlock."""
+
+import itertools
+import re
+import textwrap
+import typing
+
+import jinja2
+import pydantic
+from django.utils.translation import gettext_noop as _
+from jinja2.sandbox import SandboxedEnvironment
+from xblock.core import XBlock
+from xblock.exceptions import JsonHandlerError
+from xblock.fields import Boolean, Dict, List, Scope, String
+from xblock.validation import ValidationMessage
+from web_fragments.fragment import Fragment
+
+from .base import AIEvalXBlock
+from .llm import SupportedModels
+
+
+DEFAULT_SUPERVISOR_PROMPT = textwrap.dedent("""
+    You are a supervisor managing an interaction between the following agents: Coach, Character.
+    Based on the conversation, decide which agent should respond next.
+    You can choose from: Character, Coach, or FINISH.
+    You are responsible for managing the flow of the conversation between agents.
+    The conversation should flow naturally with the Character until specific conditions are met.
+    Switch control to the Coach only under the following conditions:
+    (1) The learner makes the same mistake **three times in a row**.
+    (2) The learner **explicitly** asks for help.
+    (3) The learner gets **significantly off topic** and is no longer addressing the learning objectives or the project.
+    If the learner shows minor deviations or uncertainty, let the Character continue interacting with the learner.
+    And if the learner specifically asks for help, you should always call on the Coach.
+    Your goal is to provide enough opportunities for the learner to self-correct and progress naturally without premature intervention.
+    Call the conversation complete and choose FINISH only under the following conditions:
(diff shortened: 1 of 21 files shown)


ai_eval/multiagent.py
```diff
@@ -0,0 +1,651 @@
...
+from xblock.fields import Boolean, Dict, List, Scope, String
+from xblock.validation import ValidationMessage
+from web_fragments.fragment import Fragment
+
+from .base import AIEvalXBlock
+from .llm import SupportedModels
+
+
+DEFAULT_SUPERVISOR_PROMPT = textwrap.dedent("""
+    You are a supervisor managing an interaction between the following agents: Coach, Character.
```
pkulkark, line 23: @ArturGaspar Was this prompt provided by WGU?

ai_eval/multiagent.py
```diff
@@ -0,0 +1,651 @@
...
+    scenario: Scenario
+    characters: ScenarioCharacters
+
+
+class Character(pydantic.BaseModel):
+    name: str
+    role: str
+
+
+class CharacterData(pydantic.BaseModel):
```
pkulkark, line 131: These pydantic models seem to be used for validation. Any reason for not using django serializers?

ai_eval/multiagent.py
```diff
@@ -0,0 +1,651 @@
...
+                user_character_name="",
+                user_character_data="",
+                scenario_data=data.scenario_data,
+            )
+        except jinja2.TemplateError as e:
+            validation.add(ValidationMessage(
+                ValidationMessage.ERROR,
+                f"{self._get_field_display_name('agent_prompt')}: {e}",
+            ))
+        else:
```
pkulkark, line 439: Why do we need `else` here? Couldn't we just continue without it?

ai_eval/multiagent.py
ArturGaspar: @pkulkark Yes.

ai_eval/multiagent.py
ArturGaspar: @pkulkark Is there any reason to prefer them?

pkulkark: @ArturGaspar Just for consistency with the rest of edx-platform xblocks. Since the xblock uses Django models, using a Django serializer would be more idiomatic.

ai_eval/templates/chatbox.html
```diff
@@ -26,10 +26,13 @@
         </div>
         <div class="submit-row">
             {% if self.allow_reset %}
-                <span id="reset-button" class="disabled-btn">Reset chat</span>
+                <span class="chat-button" id="reset-button">Reset chat</span>
             {% endif %}
-            <textarea class="user-input" rows="1" placeholder="Type your answer here" maxlength="1000"></textarea>
-            <span id="submit-button">Submit <i class="fa fa-paper-plane"></i></span>
+            {% if has_finish_button %}
+                <span class="chat-button" id="finish-button">Submit for evaluation</span>
```
pkulkark, line 32: @ArturGaspar It might be better to disable this button until there's a user response. What do you think?

ai_eval/templates/chatbox.html
```diff
@@ -26,10 +26,13 @@
...
+            {% endif %}
+            <textarea id="user-input" rows="1" placeholder="Type your answer here" maxlength="1000"></textarea>
+            <span class="chat-button" id="submit-button">Submit <i class="fa fa-paper-plane"></i></span>
```
pkulkark, line 35: Same here.

PR #10, opened 2025-03-10
feat: allow attaching files by URL to the short answer XBlock

ArturGaspar:
## Description

Allows attachments to be loaded from an URL for the short answer AI XBlock.

## Testing instructions

1. Add "shortanswer_ai_eval" and configure required settings.
2. Set "Evaluation prompt" to "Provide information on the attachments", in order to be able to ask about the attachments
3. Set `["http://example.com"]` as attachment URLs.
4. Use the XBlock, ask the LLM about the attachment contents.

image

## Other information

Private-ref: https://tasks.opencraft.com/browse/BB-9554

### DIFF
--- a/ai_eval/__init__.py
+++ b/ai_eval/__init__.py
@@ -4,3 +4,4 @@ Xblock to have short text and code entries with AI-driven evaluation.
 
 from .shortanswer import ShortAnswerAIEvalXBlock
 from .coding_ai_eval import CodingAIEvalXBlock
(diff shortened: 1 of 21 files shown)


pkulkark (approved):
Nit - Shouldn't this PR be against the `main` branch? Any specific reason for opening it against the `multiagent` branch?

Changes LGTM :+1:

- [x] I tested this: verified that the attachment url works as described in the testing instructions.
- [x] I read through the code

PR #11, opened 2025-06-18
Pooja/bb 9550 use custom service

pkulkark:
## Description

This PR adds the option to route all LLM calls through a user‑provided “custom” LLM service instead of the default provider. The option can be enabled via a feature flag named `USE_CUSTOM_LLM_SERVICE` that can be passed in through the site configuration, along with the custom service's endpoints. And the Oauth2 client credentials through Django settings.

## Supporting information

Opencraft Internal Jira task: [BB-9550](https://tasks.opencraft.com/browse/BB-9550)

## Testing instructions

1. Deploy this branch of the XBlock.
2. Enable custom service in Site Configuration. In your Open edX Django admin → Site Configurations, add:
```
           {
             "ai_eval": {
               "USE_CUSTOM_LLM_SERVICE": true,
               "CUSTOM_LLM_MODELS_URL": "https://your-custom-service/models",
               "CUSTOM_LLM_COMPLETIONS_URL": "https://your-custom-service/completions",
               "CUSTOM_LLM_TOKEN_URL": "https://your-custom-service/oauth/token"
             }
           }
```
3. Set the following client credentials in Django settings for Oauth2 (e.g. via Tutor’s configuration):
```
CUSTOM_LLM_CLIENT_ID = "your-client-id"
CUSTOM_LLM_CLIENT_SECRET = "your-client-secret"
```
4. Test the XBlock.

## Other information

Currently there's a version conflict that prevents this XBlock from being installed properly. The conflict is from the main edx-platform and documented in this [issue](https://github.com/openedx/edx-platform/issues/35268). Until that is resolved, a simple workaround could be to create a tutor plugin with the following content:
```
name: litellm-override
version: 0.1.0                                                                                                                                                                                                                                                                                                                                                                                                                                              

patches:
    openedx-dockerfile-post-python-requirements: |
        RUN pip install --no-cache-dir certifi==2023.11.17 && \
        pip install --no-cache-dir litellm==0.14.0 --no-deps
    openedx-dev-dockerfile-post-python-requirements: |
        RUN pip install --no-cache-dir certifi==2023.11.17 && \
        pip install --no-cache-dir litellm==0.14.0 --no-deps    
```

### DIFF
--- a/ai_eval/llm_services.py
+++ b/ai_eval/llm_services.py
@@ -33,9 +33,14 @@ class CustomLLMService(LLMServiceBase):
         self._expires_at     = 0
 
     def _fetch_token(self):
-        data = {'grant_type': 'client_credentials'}
-        auth = (self.client_id, self.client_secret)
-        response = requests.post(self.token_url, data=data, auth=auth)
+        data = {
+            'grant_type': 'client_credentials',
+            'client_id': self.client_id,
+            'client_secret': self.client_secret,
+            'scope': 'ml.chatbot.query'
+        }
+        headers = {'Content-Type': 'application/x-www-form-urlencoded'}
+        response = requests.post(self.token_url, data=data, headers=headers)
         response.raise_for_status()
         token_data = response.json()
         self._access_token = token_data['access_token']
--- a/ai_eval/llm_services.py
+++ b/ai_eval/llm_services.py
@@ -1,6 +1,10 @@
 import requests
+import logging
+
 from litellm import completion
 from .llm import SupportedModels
+
+logger = logging.getLogger(__name__)
 
 class LLMServiceBase:
     def get_response(self, model, api_key, messages, api_base):
@@ -77,7 +81,7 @@ class CustomLLMService(LLMServiceBase):
         response = requests.post(url, json=payload, headers=self._get_headers())
         response.raise_for_status()
         data = response.json()
-        # TODO: Adjust this if custom API returns the response differently
+        # Adjust this if custom API returns the response differently
         return data.get("response")
 
     def get_available_models(self):
@@ -94,5 +98,9 @@ class CustomLLMService(LLMServiceBase):
             elif isinstance(data, list):
                 return [str(m) for m in data]
             return []
-        except Exception:
+        except Exception as e:
+            logger.error(
+                f"Failed to fetch available models from custom LLM service. Error: {e}",
+                exc_info=True,
+            )
             return []
--- a/ai_eval/llm_services.py
+++ b/ai_eval/llm_services.py
@@ -1,0 +1,93 @@
+import requests
+from litellm import completion
+from .llm import SupportedModels
+
+class LLMServiceBase:
+    def get_response(self, model, api_key, messages, api_base):
+        raise NotImplementedError
+
+    def get_available_models(self):
+        raise NotImplementedError
+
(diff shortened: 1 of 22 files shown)


ai_eval/llm_services.py
```diff
@@ -0,0 +1,106 @@
...
+    def _fetch_token(self):
+        data = {
+            'grant_type': 'client_credentials',
+            'client_id': self.client_id,
+            'client_secret': self.client_secret,
+            'scope': 'ml.chatbot.query'
+        }
+        headers = {'Content-Type': 'application/x-www-form-urlencoded'}
+        response = requests.post(self.token_url, data=data, headers=headers)
+        response.raise_for_status()
```
tpradowgu, line 67: How these response.raise_for_status() are being managed?

ai_eval/llm_services.py
```diff
@@ -0,0 +1,106 @@
...
+        # Adjust the payload structure based on custom API requirements
+        prompt = " ".join(
+            f"{msg.get('role', '').capitalize()}: {msg.get('content', '').strip()}"
+            for msg in messages
+        )
+        payload = {
+            "model": str(model),
+            "prompt": prompt,
+        }
+        response = requests.post(url, json=payload, headers=self._get_headers())
```
tpradowgu: Is there a plan to peform retries if something goes wrong here? Same question for the other requests, but I believe here is the most important part after the user has been authenticated to perform the interaction with the xblock

tpradowgu (commented):
Perhaps I am wrong, but shouldn't the user prompts and questions fields have the scope set to Scope.user_state so the value is saved for each student and it can be retrieved later for analysis?

pkulkark:
@tpradowgu 

> Perhaps I am wrong, but shouldn't the user prompts and questions fields have the scope set to Scope.user_state so the value is saved for each student and it can be retrieved later for analysis?

The user prompts and question fields are set at the xblock level and not user level i.e. it is expected to be the same for all users. The answers submitted and the AI generated feedback are specific to the user, and those fields are set to Scope.user_state.

ai_eval/llm_services.py
pkulkark: It is wrapped within try/except in the specific xblock level where it also logs the error message.

ai_eval/llm_services.py
pkulkark: I deliberately did not enable retries since the endpoint can be hit by many users in parallel and/or the llm services could charge per requests leading to unexpected costs. If the llm service supports retries then we could add `num_retries` parameter.

ai_eval/llm_services.py
```diff
@@ -0,0 +1,121 @@
...
+        headers = {'Content-Type': 'application/x-www-form-urlencoded'}
+        response = requests.post(self.token_url, data=data, headers=headers, timeout=10)
+        response.raise_for_status()
+        token_data = response.json()
+        self._access_token = token_data['access_token']
+        expires_in = token_data.get('expires_in')
+        if expires_in:
+            self._expires_at = time.time() + expires_in - 60
+        else:
+            self._expires_at = time.time() + 3300
```
ArturGaspar: This default would be better defined as a constant.

pkulkark: Done.

rodmgwgu (approved):
Changes tested with custom LLM service, LLM selection and inference working correctly.

PR #12, opened 2025-07-17
 feat: AI XBlock conversation history export 

ArturGaspar:
## Description

XBlock to export data from the others

## Testing instructions

1. Add the setting `CELERY_EXTRA_IMPORTS = ["ai_eval.tasks"]` to deployment
2. In a course, add "shortanswer_ai_eval" to Advanced module settings
3. Add some short answer AI XBlocks, give it some input, reset chat a few times, use different users
4. In a course, add "ai_eval_export" to Advanced module settings
5. Create a new unit
6. Add the XBlock named "Data Export" 
7. Check the option to hide the unit from learners
8. Publish the unit and view it live
9. Export chat logs from the XBlock

## Deadline

"None"

## Other information

Note: this is based on artur/shortanswer-with-coach to deploy both for testing at the same time but should be rebased and merged onto main.

Private-ref: https://tasks.opencraft.com/browse/BB-9777


ai_eval/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+
+def _extract_data(block):
+    from lms.djangoapps.courseware.model_data import (
+        DjangoKeyValueStore,
+        FieldDataCache,
+    )
+
+    section_name, subsection_name, unit_name = _get_context(block)
+
+    for user in User.objects.iterator():
```
pkulkark: This will iterate through all the users in the site. When it is deployed on production, it will explode the runtime. Can you filter by course enrollment and non-staff users instead?

ai_eval/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+        "generation_time_s": generation_time_s,
+    }
+
+
+def _extract_all_data(course_id):
+    from xmodule.modulestore.django import modulestore
+
+    store = modulestore()
+    for block in store.get_items(course_id):
+        if isinstance(block, _SUPPORTED_BLOCKS):
```
pkulkark: Can this not be included as a filter while fetching the blocks? Otherwise we'll end up fetching every block of the course.

ai_eval/tasks.py
ArturGaspar: @pkulkark  I have added a filter for course enrollment. I think it makes sense to include staff users as well, so as not to exclude a conversation that a staff user had e.g. for testing.

pkulkark: Fair enough 👍

pkulkark (approved):
LGTM 👍

PR #14, opened 2025-08-11
[BB-9936] Make code execution tool configurable

pkulkark:
## Description

This PR adds a pluggable code execution backend for the Coding XBlock, allowing deployments to use either the default Judge0 service or a configurable third‑party service. The execution backend is setup via Django settings (`AI_EVAL_CODE_EXECUTION_BACKEND`). It also allows to optionally override the judge0 base URL via `judge0_conf
ig.base_url` which can be used in case of self-hosted judge0 service. If no settings are provided, existing Judge0 behaviour continues unchanged.

## Supporting information

OpenCraft's Internal Jira task: [BB-9936](https://github.com/open-craft/xblock-ai-evaluation/pull/14)

## Testing instructions

1. Check out this branch and without changing any settings, verify that the default behaviour of using judge0 rapidapi works properly.
2. For a custom service, set the following django settings through tutor plugin:
```
AI_EVAL_CODE_EXECUTION_BACKEND = { 
    'backend': 'custom',
    'custom_config': { 
        'submit_endpoint': 'https://.../submit',
        'results_endpoint': 'https://.../results/{submission_id}',
        'languages_endpoint': 'https://.../languages',
        'api_key': '<token>',
        'auth_header_name': 'Authorization',
        'auth_scheme': 'Bearer',
        'timeout': 30
    }
}
```
3. Verify that the coding xblock works as expected.

## Other information

Note: This change relies on the fact that the external service has an asynchronous model (similar to judge0) and separate `submit_endpoint` which returns a submission ID and `results_endpoint` which is polled with `{submission_id}`) for results. An additional `languages_endpoint` is used to validate supported languages.

PR #16, opened 2025-09-03
[BB-9988] Add option to use conversation_id for custom llm service

pkulkark:
## Description

This PR adds optional provider‑side threads (conversation IDs) for custom LLM services, gated by the `USE_PROVIDER_THREADS` site flag. The XBlock caches a provider `thread_id` and tracks a `thread_tag`
fingerprint (provider/model/prompt/attachments) to auto‑invalidate when settings change; Reset clears both. The shared LLM wrapper passes and persists thread IDs when supported, while the default path remains stateless.
 
Important note: The XBlock/LMS remains the canonical chat history (source of truth) so that we can preserve vendor flexibility, continuity, and deterministic behavior; provider threads are treated only as a cache and we safely fall back when unsupported.

This update affects only chat interactions; the Coding XBlock and its code‑execution flow are not impacted as they are one-time feedback only (stateless).

## Supporting information

OpenCraft Internal Jira task: [BB-9988](https://tasks.opencraft.com/browse/BB-9988)

## Screenshot

Screenshot 2025-09-04 at 2 00 48 PM

## Testing instructions

1. Deploy this branch of the XBlock.
2. Without making any config changes, verify that the default path (with the standard providers and models) is working as expected.
3. Configure a custom LLM service through the site config, make sure to include the new `USE_PROVIDER_THREADS` flag set to True.
4. Test the AI XBlocks; Both chat and coding xblocks should work without trouble.
5. To verify that the conversation_id is set correctly and used, go to django admin and check `/admin/courseware/studentmodule/` to see that the user states have a conversation_id set.

## Other information

This depends on  #14 as that contains a bunch of important code refactoring. Once #14 is merged, this will need to be rebased before merging.

### DIFF
--- a/ai_eval/base.py
+++ b/ai_eval/base.py
@@ -78,6 +78,16 @@ class AIEvalXBlock(StudioEditableXBlockMixin, XBlock):
         default="",
         values_provider=_get_model_choices,
     )
+    thread_id = String(
+        help=_("Provider conversation/thread identifier (cache only)"),
+        default="",
+        scope=Scope.user_state,
+    )
+    thread_tag = String(
+        help=_("Fingerprint of provider/model/prompt for thread reuse"),
+        default="",
+        scope=Scope.user_state,
+    )
 
     editable_fields = (
         "display_name",
@@ -207,5 +217,17 @@ class AIEvalXBlock(StudioEditableXBlockMixin, XBlock):
                 )
 
     def get_llm_response(self, messages):
-        return get_llm_response(self.model, self.get_model_api_key(), messages,
-                                self.get_model_api_url())
+        """
+        Call the shared LLM entrypoint and return the response.
+        """
+        prior_thread_id = self.thread_id or None
+        text, new_thread_id = get_llm_response(
+            self.model,
+            self.get_model_api_key(),
+            messages,
+            self.get_model_api_url(),
+            thread_id=prior_thread_id,
+        )
+        if new_thread_id:
+            self.thread_id = new_thread_id
+        return text
--- a/ai_eval/base.py
+++ b/ai_eval/base.py
@@ -6,7 +6,7 @@ import pkg_resources
 
 from django.utils.translation import gettext_noop as _
 from xblock.core import XBlock
-from xblock.fields import String, Scope
+from xblock.fields import String, Scope, Dict
 from xblock.utils.resources import ResourceLoader
 from xblock.utils.studio_editable import StudioEditableXBlockMixin
 from xblock.validation import ValidationMessage
@@ -78,14 +78,9 @@ class AIEvalXBlock(StudioEditableXBlockMixin, XBlock):
         default="",
         values_provider=_get_model_choices,
     )
-    thread_id = String(
-        help=_("Provider conversation/thread identifier (cache only)"),
-        default="",
-        scope=Scope.user_state,
-    )
-    thread_tag = String(
-        help=_("Fingerprint of provider/model/prompt for thread reuse"),
-        default="",
+    thread_map = Dict(
+        help=_("Map of provider thread IDs keyed by tag"),
+        default={},
         scope=Scope.user_state,
     )
 
@@ -216,11 +211,17 @@ class AIEvalXBlock(StudioEditableXBlockMixin, XBlock):
                     )
                 )
 
-    def get_llm_response(self, messages):
+    def get_llm_response(self, messages, tag: str | None = None):
         """
-        Call the shared LLM entrypoint and return the response.
+        Call the shared LLM entrypoint and return only the response text.
         """
-        prior_thread_id = self.thread_id or None
+        prior_thread_id = None
+        if tag:
+            try:
+                prior_thread_id = (self.thread_map or {}).get(tag) or None
+            except Exception:  # pylint: disable=broad-exception-caught
+                prior_thread_id = None
+
         text, new_thread_id = get_llm_response(
             self.model,
             self.get_model_api_key(),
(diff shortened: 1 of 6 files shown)


ai_eval/llm_services.py
```diff
@@ -81,25 +117,56 @@ def _get_headers(self):
...
+            thread_id=None,
+            use_threads=False
+    ):
         """
         Send completion request to custom LLM endpoint.
         """
         url = self.completions_url
+        # When reusing an existing thread, only send the latest user input and rely on
+        # the provider to apply prior context associated with the conversation_id.
+        if use_threads and thread_id:
```
xitij2000: I'm not sure what purpose the `use_threads` flag is fulfilling. 

It would be derived from `supports_threads` but the LLM Service already knows the value? So why are we calling a function on the LLM service to get this value and then passing it back to the LLM service? 

For most cases a None value for thread_id will suffice. The only use `use_threads` has, is that it creates a new thread.

I think it might be better to segregate creation of a new thread since it will support more rich scenarios such as having multiple conversation threads. 

i.e. if in the future this block wants to support creating multiple threads for each user.

So the calling function flow can then be:
if llm.supports_threads:
   thread_id = llm.new_thread()
llm. get_response(...)

ai_eval/llm_services.py
```diff
@@ -153,3 +220,10 @@ def get_available_models(self):
                 exc_info=True,
             )
             return []
+
+    def supports_threads(self) -> bool:
```
xitij2000, line 259: Wouldn't this be better as a block-level option? 

i.e. not in site-configuration but in the block itself, it can be defaulted to true.

ai_eval/shortanswer.py
```diff
@@ -178,15 +181,36 @@ def get_response(self, data, suffix=""):  # pylint: disable=unused-argument
...
+        prompt_hash = prompt_hasher.hexdigest()
+
+        # Determine provider tag based on service type
+        llm_service = get_llm_service()
+        provider_tag = "custom" if isinstance(llm_service, CustomLLMService) else "default"
+        current_tag = f"{provider_tag}:{self.model}:{prompt_hash}"
+        # Invalidate stale threads when tag changes
+        if getattr(self, "thread_tag", "") != current_tag:
+            self.thread_id = ""
+
```
xitij2000: Is storing conversation history a chargeable thing? If so we should probably clean it up when no longer in use.

ai_eval/base.py
```diff
@@ -78,6 +78,16 @@ class AIEvalXBlock(StudioEditableXBlockMixin, XBlock):
         default="",
         values_provider=_get_model_choices,
     )
+    thread_id = String(
+        help=_("Provider conversation/thread identifier (cache only)"),
+        default="",
+        scope=Scope.user_state,
+    )
```
xitij2000: Would it make sense to make this a dict? That way, in the future, if needed, you can support multiple threads.

xitij2000 (commented):
Generally everything looks good from a code perspective. Just have a few suggestions

ai_eval/llm_services.py
pkulkark: The reason `supports_threads` is part of LLM Service is because it's dependent on the particular service. Most standard services that use API based requests, are stateless and don't support threads. That's why it made more sense to add it into the `CustomLLMService` and keep it as a site config option.

ai_eval/llm_services.py
pkulkark: `use_threads` is used as the per-call directive whereas `supports_threads()` is the capability/policy gate. So should there ever be a need to force stateless behaviour for a single call, we can just pass `use_threads=False`.

xitij2000: I understand that that is what it does, however, if you trace the usage of use_threads and thread_id there is scope for simplification:

| use_threads  | thread_id | action |
|--------------------|----------------|-----------|
| False | Falsy   | non-threaded conversation |
| False | Truthy | non-threaded conversation |
| True  | Falsy    | start new thread |
| True  | Truthy | threaded conversation |

So if use_threads if false, then sending a thread_id has no impact. So I feel it could be reorganised into two paths:

1. Create a new thread in a separate function. 
2. If a thread_id is provided, the conversation is threaded, if not, it's non threaded, no need to gate it with use_threads, since you can just choose to omit thread_id and get the same result. 


I feel it would be cleaner that way since right now the function is doing too much.

ai_eval/llm_services.py
```diff
@@ -81,25 +117,56 @@ def _get_headers(self):
...
         # Adjust this if custom API returns the response differently
-        return data.get("response")
+        text = data.get("response")
+        # Try to capture a new conversation/thread id if the API returns one.
+        new_thread_id = None
+        if use_threads:
+            new_thread_id = (
+                data.get("conversation_id")
+                or (data.get("data", {}) if isinstance(data.get("data"), dict) else {}).get("conversation_id")
+            )
```
xitij2000: I think this code is a bit hard to follow, it would be better to expand it a bit to make it easier to follow:


```suggestion
        if use_threads:
            new_thread_id = data.get("conversation_id")
            if not new_thread_id and isinstance(data.get("data"), dict):
                new_thread_id = data["data"].get("conversation_id")
```

ai_eval/llm_services.py
xitij2000: I think in that case perhaps this config could be called `PROVIDER_SUPPORTS_THREADS` instead of `USE_PROVIDER_THREADS` since the latter seems to imply that config is about whether to use the feature or not.

xitij2000 (approved):
I think the code is looking good! I just had a few suggestions. 

I think for the splitting out of the create_thread functionality, it seems like a good approach for me, however you've worked on this a lot more and have a deeper undersanding of the code. So if you feel that my suggestion doesn't make sense or is infeasible, then you can ignore it.

ai_eval/llm_services.py
pkulkark: @xitij2000 You're right, that would be much cleaner. Thanks for pointing that out. I've updated the code with your suggestions.

ai_eval/llm_services.py
pkulkark: Done

ai_eval/llm_services.py
pkulkark: Updated.

xitij2000 (approved):
Looks good!

PR #21, opened 2026-01-08
[BB-10439] Update AI Export Script

pkulkark:
## Description

This PR updates the AI conversation history export script as follows:
1. Course title is added as the first column of the report
2. Multiagent XBlock data is replaced with Coach AI XBlock data

## Supporting information

OpenCraft Internal Jira task: [BB-10439](https://tasks.opencraft.com/browse/BB-10439)

## Testing instructions

1. Install this branch of the XBlock.
2. Interact with a few of the AI XBlocks (short answer, coding and coaching).
3. Login as staff and add `ai_eval_export` in the advanced module list and add it anywhere in the course and publish.
4. Open the published LMS page as staff and generate and download the report.
5. Verify that the report contains the course name as the first column and it includes the coaching xblock conversation history.

### DIFF
--- a/ai_eval/tasks.py
+++ b/ai_eval/tasks.py
@@ -20,11 +20,70 @@ logger = get_task_logger(__name__)
 
 User = get_user_model()
 
+_BASE_HEADER = (
+    "Section",
+    "Subsection",
+    "Unit",
+    "Location",
+    "Display Name",
+    "Username",
+    "User E-mail",
+    "Conversation",
+    "Source",
+    "Message",
+)
+
 _BLOCK_CATEGORIES = [
     'coding_ai_eval',
     'multiagent_ai_eval',
     'shortanswer_ai_eval',
 ]
+
+
+def _get_course_display_name(course_id):
+    """
+    Return a human-readable course title for `course_id`.
+
+    Best-effort: on any error, fall back to `str(course_id)`.
+    """
+    for getter in (
+        _get_course_display_name_from_course_overview,
+        _get_course_display_name_from_modulestore,
+    ):
+        try:
+            display_name = getter(course_id)
+        except Exception:  # pylint: disable=broad-exception-caught
+            display_name = None
+        if display_name:
+            return str(display_name)
+    return str(course_id)
+
+
+def _get_course_display_name_from_course_overview(course_id):
+    # pylint: disable=import-error,import-outside-toplevel
+    from openedx.core.djangoapps.content.course_overviews.models import CourseOverview
+
+    overview = CourseOverview.get_from_id(course_id)
+    return getattr(overview, "display_name", None)
+
+
+def _get_course_display_name_from_modulestore(course_id):
+    # pylint: disable=import-error,import-outside-toplevel
+    from xmodule.modulestore.django import modulestore
+
+    course = modulestore().get_course(course_id)
+    return getattr(course, "display_name", None) or getattr(
+        course, "display_name_with_default", None
+    )
+
+
+def _build_export_rows(course_display_name, data_rows_iter):
+    """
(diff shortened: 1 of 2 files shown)


ai_eval/tasks.py
```diff
@@ -12,21 +12,86 @@
...
+    return getattr(course, "display_name", None) or getattr(
+        course, "display_name_with_default", None
+    )
+
+
+def _build_export_rows(course_display_name, data_rows_iter):
+    """
+    Build an export row iterator, prefixing each row with `course_display_name`.
+    """
+    header = ("Course Name",) + _BASE_HEADER
```
samuelallan72: nit: "Course Name" can be part of the header constant.

ai_eval/tasks.py
```diff
@@ -12,21 +12,86 @@
...
 
+def _get_course_display_name(course_id):
+    """
+    Return a human-readable course title for `course_id`.
+
+    Best-effort: on any error, fall back to `str(course_id)`.
+    """
+    for getter in (
+        _get_course_display_name_from_course_overview,
+        _get_course_display_name_from_modulestore,
```
samuelallan72, line 52: do we really need this fallback logic? Under what circumstances would one or the other fail?

ai_eval/tasks.py
```diff
@@ -93,18 +155,63 @@ def _get_messages(block, session):
...
+    Yield (source, content) tuples for CoachAIEvalXBlock user state.
+
+    Ordering is best-effort: workspace first, then coach, then evaluation fragments.
+    """
+    main_role = getattr(block, "character_1_role", None) or getattr(
+        block, "character_1_name", None
+    ) or "Main character"
+    coach_role = getattr(block, "character_2_role", None) or getattr(
+        block, "character_2_name", None
+    ) or "Coach"
```
samuelallan72: Why do we need to be so defensive here? These attributes should always exist on the xblock right?

ai_eval/tasks.py
```diff
@@ -140,10 +250,35 @@ def _extract_data(block):
                     block.display_name,
                     user.username,
                     user.email or "",
-                    idx,
+                    1,
```
samuelallan72: Does this mean it won't be possible to filter by conversation?

I guess you can still filter by (location, username) to get a specific conversation. But maybe we shouldn't hardcode the conversation ID here if we can't get one; perhaps set it to None instead? Otherwise it could be misleading.

samuelallan72 (commented):
I tested this on the sandbox at it appears to work fine. 👍  I left some code quality nits to address as you see fit. On the documentation side, it would be nice to use type annotations more. :)

The main thing I flagged that I think needs to be addressed as the hardcoding of conversation id for the coach rows.

- [x] I tested this: tested on the sandbox (thanks for deploying to the sandbox for testing!)
- [x] I read through the code
- [x] Includes documentation

samuelallan72 (approved):

ai_eval/tasks.py
pkulkark: CourseOverview is a cached representation and may be missing/outdated. I noticed the "CourseOverview first, then modulestore” idea in a couple of other places (like [here](https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/content/course_overviews/models.py#L382)) and went with the same.

ai_eval/tasks.py
pkulkark: You're right. I got a little carried away with it 😄 Will fix it.

ai_eval/tasks.py
pkulkark: Good point. I've set it to None.

ai_eval/tasks.py
pkulkark: Done.

PR #23, opened 2026-02-23
fix: Move judge0 API key to config and disable studio API-keys

pkulkark:
## Description

This PR updates the AI Evaluation Studio flow so self-hosted Judge0 API keys are now read from platform configuration instead of relying on per-XBlock field values. It also locks API key fields in Studio when the required keys are already configured at the site/global level.

## Supporting information

OpenCraft Internal Jira task - https://tasks.opencraft.com/browse/BB-10538

## Testing instructions

1. Deploy this branch.
2. Without any site config changes, verify that the model api key field is editable for any of the XBlock (coding or short answer or coaching).
3. Add a model api key in the site config as described [here](https://github.com/open-craft/xblock-ai-evaluation/tree/pooja/bb10538-config-judge0-api-key?tab=readme-ov-file#api-configuration) in the README.
4. In the xblock settings editor, chose the corresponding model from the dropdown and verify that the model api key field is locked (not editable).
5. Go back to the site config, remove the api key and set `"USE_CUSTOM_LLM_SERVICE": true`.
6. Since we haven't completed the custom llm service setup, the xblock settings will fail to save but you can verify that the model api key field is locked.
7. Set Judge0 configuration as described [here](https://github.com/open-craft/xblock-ai-evaluation/tree/pooja/bb10538-config-judge0-api-key?tab=readme-ov-file#custom-code-execution-service-advanced) in the README.
8. In the coding XBlock editor, verify that the Judge0 api key field is also locked (not editable).

### DIFF
--- a/ai_eval/backends/factory.py
+++ b/ai_eval/backends/factory.py
@@ -14,18 +14,16 @@ class BackendFactory:
         """
         Get the appropriate backend based on Django settings.
 
-        Args:
-            api_key: Judge0 API key (only used for judge0 backend)
-
         Returns:
             CodeExecutionBackend: Configured backend instance
         """
-        backend_config = getattr(
-            django_conf.settings, 'AI_EVAL_CODE_EXECUTION_BACKEND', {}
+        raw_backend_config = getattr(
+            django_conf.settings, 'AI_EVAL_CODE_EXECUTION_BACKEND', None
         )
+        has_backend_config = raw_backend_config is not None
 
-        if backend_config.get('backend') == 'custom':
-            config = backend_config.get('custom_config', {})
+        if has_backend_config and raw_backend_config.get('backend') == 'custom':
+            config = raw_backend_config.get('custom_config', {})
             return CustomServiceBackend(
                 submit_endpoint=config.get('submit_endpoint', ''),
                 results_endpoint=config.get('results_endpoint', ''),
@@ -37,8 +35,17 @@ class BackendFactory:
             )
 
         # Default to judge0 backend
-        judge0_config = backend_config.get('judge0_config', {})
+        judge0_config = (
+            raw_backend_config.get('judge0_config', {})
+            if has_backend_config
+            else {}
+        )
+        judge0_api_key = (
+            judge0_config.get('api_key', '')
+            if has_backend_config
+            else api_key
+        )
         return Judge0Backend(
-            api_key=api_key,
+            api_key=judge0_api_key,
             base_url=judge0_config.get('base_url')
         )
--- a/ai_eval/backends/factory.py
+++ b/ai_eval/backends/factory.py
@@ -17,6 +17,9 @@ class BackendFactory:
         Returns:
             CodeExecutionBackend: Configured backend instance
         """
+        # Keep `None` as the default so we can distinguish "setting absent"
+        # from "setting present but incomplete". We only fall back to the
+        # per-block `api_key` when the backend setting is truly absent.
         raw_backend_config = getattr(
             django_conf.settings, 'AI_EVAL_CODE_EXECUTION_BACKEND', None
         )
--- a/ai_eval/base.py
+++ b/ai_eval/base.py
@@ -150,7 +150,10 @@ class AIEvalXBlock(StudioEditableXBlockMixin, XBlock):
 
         if config_parameter == "api_key":
             # When configured globally/site-wide, ignore XBlock-local model_api_key.
-            if value := self._get_site_or_global_model_config_value(config_parameter, obj):
+            if value := self._get_site_or_global_config_value(
+                config_parameter,
+                obj=obj,
+            ):
                 return value
             if value := getattr(obj, field_name, None):
                 return str(value)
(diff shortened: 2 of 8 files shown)


ai_eval/backends/factory.py
```diff
@@ -14,18 +14,16 @@ def get_backend(cls, api_key: str = ""):
...
-            api_key: Judge0 API key (only used for judge0 backend)
-
         Returns:
             CodeExecutionBackend: Configured backend instance
         """
-        backend_config = getattr(
-            django_conf.settings, 'AI_EVAL_CODE_EXECUTION_BACKEND', {}
+        raw_backend_config = getattr(
+            django_conf.settings, 'AI_EVAL_CODE_EXECUTION_BACKEND', None
         )
```
farhaanbukhsh, line 25: @pkulkark I think this is an anti-pattern that we should clean.


```suggestion
        from django.conf import settings
        raw_backend_config = settings.get('AI_EVAL_CODE_EXECUTION_BACKEND', None)
```

We should follow the above pattern unless there is a reason why we are doing things the way we are now. In that case, we should add a comment about it.

What do you think?

ai_eval/backends/factory.py
```diff
@@ -37,8 +35,17 @@ def get_backend(cls, api_key: str = ""):
             )
 
         # Default to judge0 backend
-        judge0_config = backend_config.get('judge0_config', {})
+        judge0_config = (
+            raw_backend_config.get('judge0_config', {})
+            if has_backend_config
```
farhaanbukhsh, line 43: if we change the below line to have `{}` instead of None

raw_backend_config = settings.get('AI_EVAL_CODE_EXECUTION_BACKEND', {})

Now we do not have to do a `has_backend_config = raw_backend_config is not None` because calling get on an empty dict will give us None, hence we can do all the get calls normally here.

Please let me know if that doesn't sound right to you.

ai_eval/static/js/src/studio_api_key_lock.js
```diff
@@ -0,0 +1,99 @@
...
+            return modelField.value;
+        }
+        return (data && data.initial_model) || "";
+    };
+
+    var isModelApiKeyLocked = function(model) {
+        if (useCustomService) {
+            return true;
+        }
+        if (Object.prototype.hasOwnProperty.call(modelKeyPresence, model)) {
```
farhaanbukhsh:
```suggestion
        if (modelKeyPresence.hasOwnProperty(model)) {
```
this should do the job here @pkulkark, since we are already giving this a dictionary or an empty one.

ai_eval/base.py
```diff
@@ -144,25 +148,124 @@ def _get_model_config_value(self, config_parameter: str, obj: Self = None) -> st
...
+        if value := get_site_configuration_value(self.block_settings_key, config_key):
+            return value
+        return self._get_settings().get(config_key)
+
+    @staticmethod
+    def _get_model_config_key(model: str, config_parameter: str) -> str:
+        """Build model configuration key name for site/global settings lookups."""
         # For custom models, use the model name directly; for supported models, use the enum name
         try:
-            model_name = SupportedModels(obj.model).name
```
farhaanbukhsh, line 149: I am so happy to see this 👍🏾

ai_eval/coding_ai_eval.py
```diff
@@ -164,6 +163,32 @@ def author_view(self, context=None):
 
         return self.student_view(context=context)
 
+    def _get_code_execution_backend_config(self):
+        """Return code execution backend config from Django settings, or None if absent."""
+        return getattr(settings, 'AI_EVAL_CODE_EXECUTION_BACKEND', None)
```
farhaanbukhsh, line 168: same as above!

ai_eval/base.py
```diff
@@ -144,25 +148,124 @@ def _get_model_config_value(self, config_parameter: str, obj: Self = None) -> st
...
-            return str(value)
+    def _get_site_or_global_model_config_value(
+        self, config_parameter: str, obj: Self = None
+    ) -> str | None:
+        """
+        Return model config from site or global settings only (ignores XBlock-local fields).
+        """
+        obj = obj or self
+        return self._get_site_or_global_model_config_value_for_model(
+            getattr(obj, "model", None), config_parameter
```
farhaanbukhsh:
```suggestion
            self.model, config_parameter
```

will this not work? Do we really need to pass obj?

ai_eval/base.py
```diff
@@ -144,25 +148,124 @@ def _get_model_config_value(self, config_parameter: str, obj: Self = None) -> st
...
+        )
 
-        # Site configuration
+    def _get_site_or_global_model_config_value_for_model(
```
farhaanbukhsh: Can we rename this function and the one above ? The name gets confusing.

farhaanbukhsh:
```suggestion
    def _config_value_for_model(
```

farhaanbukhsh (changes requested):
@pkulkark I have reviewed the PR and added a few questions and comments. I also think that we should carve out some time after delivering this fix to refactor it.

ai_eval/backends/factory.py
pkulkark: @farhaanbukhsh In Django, settings is a LazySettings object, not a dict, so `settings.get(...)` results in `AttributeError: 'Settings' object has no attribute 'get'`. 
The `getattr` method is the most commonly used way in other Xblocks too. See [example1](https://github.com/openedx/edx-ora2/blob/master/openassessment/fileupload/backends/s3.py#L75) and [example2](https://github.com/open-craft/xblock-poll/blob/master/poll/poll.py#L393).

ai_eval/backends/factory.py
pkulkark: This was done to differentiate between "settings absent" and "setting present but incomplete". We want it to default to the per-xblock api key only if the settings are absent.

ai_eval/base.py
pkulkark: `obj` is the draft form data used during `validate_field_data`, so `obj.model` can differ from `self.model` before save.

ai_eval/static/js/src/studio_api_key_lock.js
pkulkark: Updated.

ai_eval/base.py
pkulkark: I removed one of them and simplified the name.

ai_eval/backends/factory.py
farhaanbukhsh: @pkulkark thanks, that is a learning for me :)

ai_eval/base.py
farhaanbukhsh: Thank you so much. This feels so much better.

ai_eval/backends/factory.py
farhaanbukhsh: I think adding this explanation as a comment would help others working on the codebase.

farhaanbukhsh (approved):
👍

I have added a trivial comment but overall this looks great. Thanks for the work here.

✅ I tested this:
- by installing it in the dev stack
- check if the API_KEY is set in sit config it is being locked on the xblock config
- check if we use custom llms then the api key is be default locked.
✅ I read through the code
✅   I checked for accessibility issues
✅   Includes documentation

ai_eval/backends/factory.py
pkulkark: Added.

PR #24, opened 2026-03-30
[BB-10599] AI XBlocks redesign

pkulkark:
## Description

 Redesign of the AI XBlocks (Short Answer, Coding, Coaching) covering the following main areas:

- Multiagent XBlock removed.
- React migration: All three XBlock types migrated from jQuery/vanilla JS to React + TypeScript with shared plumbing (webpack, react-intl, common components and utilities).
- Paragon integration: Replace custom components and styles with Paragon components wherever possible.
- Coaching XBlock redesign based on figma.

## Supporting information

OpenCraft Internal Jira task: [BB-10599](https://tasks.opencraft.com/browse/BB-10599)

## Testing instructions

- Deploy branch to Tutor devstack and add all three XBlock types to a unit.
- Verify short answer XBlock - configure xblock in studio editor and verify learner view, confirm everything works as expected.
- Verify coding XBlock - configure in studio editor and verify learner view, confirm everything works as expected.
- Verify Coaching XBlock matches the new design in figma. Confirm learner and studio view works as expected.

## Other information

Note: This is a fairly large refactor.


ai_eval/static/css/coach_studio.css
```diff
@@ -1,53 +1,475 @@
...
+    --coaching-nav-active: #e5e5e5;
+    --coaching-text: #3c3c3c;
+    --coaching-text-muted: #717171;
+    --coaching-title: #515151;
+    --coaching-accent: #0078b1;
+    --coaching-accent-soft: #dceef6;
+    --coaching-danger: #a11d1d;
+    --coaching-warning-bg: #fff8e5;
+    --coaching-warning-border: #ead9a4;
+}
```
xitij2000, line 18: I don't think these should be hardcoded. If you use paragaon variable names then this can potentially be automatically themed.

ai_eval/static/css/shortanswer_studio.css
xitij2000, line 1: A couple of points:
- You can consider nesting this to make it easier to follow.
- Instead of repeated styling like "padding: 0" etc you can use the classes like "p-0", "m-0", "pt-0"
- Remove some of the hardcoded values in favor of paragon colors. Even otherwise you can use a token variable like `--pgn-color-ai-eval-xblock-color` or similar to get support for theming.

ai_eval/coach.py
```diff
@@ -106,11 +107,7 @@ class CoachAIEvalXBlock(AIEvalXBlock):
 
     evaluator_prompt = String(
         display_name=_("Evaluator prompt"),
-        help=_(
-            "Prompt used to instructs the model how to evaluate learners. "
-            "You can use Jinja variables (e.g. scenario_data.evaluation_criteria). "
-            "Learn more: https://jinja.palletsprojects.com/en/stable/templates/"
-        ),
+        help=_(""),
```
xitij2000, line 158: Why is this being removed?

ai_eval/coach.py
```diff
@@ -355,13 +352,25 @@ class CoachAIEvalXBlock(AIEvalXBlock):
...
-            fragment.add_css(self.resource_string("static/css/coach_studio.css"))
-        except Exception:  # pylint: disable=broad-exception-caught
-            pass
+    def studio_view(self, context=None):
+        """Render the React Studio editor for Coaching."""
+        fragment = Fragment('<div data-ai-eval-react-root="true"></div>')
+        fragment.add_css(self.resource_string("static/bundles/shared.css"))
+        fragment.add_css(self.resource_string("static/css/studio_api_key_lock.css"))
+        fragment.add_css(self.resource_string("static/css/shortanswer_studio.css"))
+        fragment.add_css(self.resource_string("static/css/coach_studio.css"))
```
xitij2000: Since these are generated assets and all need to be used, is there any advantage in having multiple files vs one single css? 

Also, you consider using `add_css_url` instead of `add_css` since the latter will just dump the whole css into the page.

ai_eval/coding_ai_eval.py
```diff
@@ -116,18 +116,15 @@ def student_view(self, context=None):
...
-        html = self.loader.render_django_template(
-            "/templates/coding_ai_eval.html",
-            {
-                "self": self,
-            },
+        frag = Fragment(
+            self.loader.render_django_template(
+                "/templates/coding_ai_eval.html",
+                {},
+            )
```
xitij2000: You can use `self.resource_string` since there is no django template to render in that file.

frontend/src/coaching/CoachingStudentApp.tsx
```diff
@@ -0,0 +1,1073 @@
...
+import {
+  CoachingAttemptState,
+  CoachingCharacter,
+  CoachingFinalReport,
+  CoachingMessage,
+  CoachingStudentPayload,
+} from "./types";
+
+type PaneKey = "workspace" | "coach";
+type CoachMode = "chat" | "report" | "review";
```
xitij2000, line 17: nit: Enums might be better for this.

frontend/src/coaching/CoachingStudentApp.tsx
```diff
@@ -0,0 +1,1073 @@
...
+                    defaultMessage: "Avatar for {name}",
+                  },
+                  { name: character.name },
+                )
+              : intl.formatMessage({
+                  id: "coaching.student.avatar",
+                  defaultMessage: "Avatar",
+                })
+          }
+        />
```
xitij2000: This should move to messages.ts. Additionally, you can use a single formatMessage for this:
```
{ 
  id: "...", 
  defaultMessage: ` {name, select, null {User Avatar} other {Avatar for {name}} ` 
}
```
This will output "User Avatar" when the name is null and "Avatar for {name}" when it isn't null.

frontend/src/coaching/CoachingStudentApp.tsx
```diff
@@ -0,0 +1,1073 @@
...
+  }
+
+  return <div className="coach-message__avatar coach-message__avatar--empty" aria-hidden="true" />;
+}
+
+function ChatMessage({ message }: { message: PendingCoachingMessage }) {
+  const pane = message.pane === "coach" ? "coach" : "workspace";
+  const isUser = Boolean(message.is_user);
+  const character = normalizeCharacter(message.character, pane);
+  const className =
```
xitij2000: Consider using template strings for this. Or the classnames library.

frontend/src/coaching/coachingStudioAdapters.ts
```diff
@@ -0,0 +1,211 @@
...
+  try {
+    return JSON.parse(rawValue);
+  } catch (error) {
+    return fallback;
+  }
+}
+
+function asJsonObject(rawValue: unknown): JsonObject {
+  const parsedValue =
+    typeof rawValue === "string" ? parseJsonValue(rawValue, {}) : rawValue;
```
xitij2000: parseJsonValue is already performing this string type check. 

Also AFAIK, you don't need to check if it's a string. If it's not a string and can't be parsed as JSON it'll throw and error and return the fallback.

ai_eval/base.py
```diff
@@ -264,92 +280,259 @@ def _get_studio_lock_payload(self) -> dict:
...
+    ) -> dict[str, Any]:
+        """
+        Return the normalized view payload shape used by future views.
+        """
+        return {
+            "view": view,
+            "handler_urls": handler_urls,
+            "initial_state": initial_state,
+            "meta": meta,
+        }
```
xitij2000, line 479: This isn't really doing much? It's just putting the params into a dict, which you can do by just calling dict(*params)

ai_eval/base.py
```diff
@@ -264,92 +280,259 @@ def _get_studio_lock_payload(self) -> dict:
...
+
+    def _handler_url(self, handler_name: str) -> str:
+        """
+        Build a handler URL for view payloads.
+        """
+        return self.runtime.handler_url(self, handler_name)
```
xitij2000: Likewise, this call can be replaced by simply calling self.runtime.handler_url

ai_eval/base.py
```diff
@@ -264,92 +280,259 @@ def _get_studio_lock_payload(self) -> dict:
...
+
+    @staticmethod
+    def _studio_submit_response(
+        success: bool,
+        validation_errors: dict[str, Any] | None = None,
```
xitij2000, line 484: AFAIK if you're using dict[str, Any] you can omit the Any and just use dict[str]

ai_eval/coach.py
```diff
@@ -355,13 +352,25 @@ class CoachAIEvalXBlock(AIEvalXBlock):
...
+        fragment.initialize_js(
+            "CoachAIEvalXBlockStudio",
+            self._build_view_payload(
+                view="studio",
+                handler_urls={
+                    "studio_submit": self._handler_url("studio_submit"),
+                },
+                initial_state=self._studio_initial_state(),
+                meta=self._studio_payload_meta(),
+            ),
```
xitij2000, line 420:
```suggestion
            dict(
                view="studio",
                handler_urls={
                    "studio_submit": self._handler_url("studio_submit"),
                },
                initial_state=self._studio_initial_state(),
                meta=self._studio_payload_meta(),
            ),
```

ai_eval/coach.py
```diff
@@ -370,37 +379,146 @@ def _render_template(self, template, **context):
...
+            return "scenario_evaluation_criteria"
+        return "scenario_data"
+
+    def _format_scenario_data_validation_error(
+        self,
+        error: dict[str, typing.Any],
+    ) -> tuple[str, str]:
+        """Return a Studio field key and author-facing validation message."""
+        location = tuple(error.get("loc", ()))
+        mapped_field = self._map_scenario_data_validation_error_field(location)
```
xitij2000, line 453: I'm not seeing the point of this mapping. Why not just perform the same check int eh calling function? Why convert "case_details" to "scenario_case_details" and then check for that value when you can just check for the original value? Am I missing something?

ai_eval/coach.py
```diff
@@ -370,37 +379,146 @@ def _render_template(self, template, **context):
...
+                ).format(index=location[1] + 1)
+            return mapped_field, _(
+                "Evaluation criteria must be a list of criteria with names."
+            )
 
+        return mapped_field, _(
+            "Scenario data structure is invalid. Expected keys: "
+            "case_details (str), learning_objectives (list[str]), "
+            "evaluation_criteria (list[{name: str}])."
+        )
```
xitij2000, line 491: Pydantic already has powerful validation features. I think you should move all this to validators there.

frontend/src/coaching/CoachingStudioApp.tsx
```diff
@@ -0,0 +1,1613 @@
...
+  }
+
+  try {
+    return JSON.stringify(value, null, 2);
+  } catch (error) {
+    return "";
+  }
+}
+
+function normalizeInitialState(initialState: CoachingStudioState): CoachingStudioState {
```
xitij2000: Why is this needed? The server should already provide validated state. 
Additionally, it isn't very DRY.  You can create a like `stringOrDefault(val, default) => (typeof val ==="string") ? val : default` and apply that to all the keys that need to be strings.

frontend/src/coaching/coachingStudioAdapters.ts
```diff
@@ -0,0 +1,211 @@
...
+}
+
+export function updateScenarioDataValue(
+  rawScenarioData: unknown,
+  fieldName: "case_details",
+  nextValue: string,
+) {
+  const scenarioData = {
+    ...asJsonObject(rawScenarioData),
+  };
```
xitij2000: I don't think we need a copy of the object here, it was newly created.

frontend/src/coaching/CoachingStudioApp.tsx
```diff
@@ -0,0 +1,1613 @@
...
+    max_attempts:
+      initialState.max_attempts === null || typeof initialState.max_attempts === "undefined"
+        ? ""
+        : initialState.max_attempts,
+    model: typeof initialState.model === "string" ? initialState.model : "",
+    model_api_key:
+      typeof initialState.model_api_key === "string" ? initialState.model_api_key : "",
+    model_api_url:
+      typeof initialState.model_api_url === "string" ? initialState.model_api_url : "",
+    scenario_data: stringifyJsonValue(initialState.scenario_data || {}),
```
xitij2000: Why is the JSON being stringified? I see later that it's being converted back to an object.

frontend/src/coaching/coachingStudioAdapters.ts
pkulkark: Good point. Removed copying.

frontend/src/coaching/coachingStudioAdapters.ts
pkulkark: Removed the redundant check.

frontend/src/coaching/CoachingStudentApp.tsx
pkulkark: Good idea. Updated it.

frontend/src/coaching/CoachingStudentApp.tsx
pkulkark: Updated.

ai_eval/static/css/coach_studio.css
pkulkark: Moved some fields to Paragon components. Still some CSS overrides were needed to preserve the coaching editor's custom visual design on top of Paragon's defaults and values that don't map to Paragon tokens.

ai_eval/static/css/shortanswer_studio.css
pkulkark: Same thing here - moved to Paragon components and kept some CSS overrides to match custom design.

ai_eval/coach.py
pkulkark: This was changed in the new design. The react frontend now provides its own inline field descriptions directly in the JSX.

ai_eval/coach.py
pkulkark: Done. But had to keep view-specific CSS files via add_css due to specific overrides needed.

ai_eval/coding_ai_eval.py
pkulkark: Done.

frontend/src/coaching/CoachingStudentApp.tsx
pkulkark: Apparently TypeScript string unions ("workspace" | "coach") are the recommended pattern over enums as there's no runtime code generated.

ai_eval/base.py
pkulkark: It is to provide a consistent interface for building payloads across all three XBlock types.

ai_eval/base.py
pkulkark: Done.

ai_eval/coach.py
pkulkark: Yeah this is bit of a nuance with this redesign. The new design presents `case_details`, `learning_objectives` and `evaluation_criteria` as separate form fields for better authoring UX. But the backend has `scenario_data` as a single field with all of them as dict fields. Splitting it in the backend is not preferred as this might be running in production and I wanted to avoid migrations.

ai_eval/coach.py
pkulkark: Done.

frontend/src/coaching/CoachingStudioApp.tsx
pkulkark: Good catch. Fixed it.

frontend/src/coaching/CoachingStudioApp.tsx
pkulkark: Fixed it.

ai_eval/coach.py
xitij2000: I think these should still be kept in place for anyone reading the code.

.nvmrc
```diff
@@ -0,0 +1 @@
+18
```
xitij2000, line 1: The current version used by the platform is 24.

frontend/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "typecheck": "tsc --noEmit",
+    "watch": "webpack --mode development --watch"
+  },
+  "dependencies": {
+    "@openedx/paragon": "^22.8.1",
+    "dompurify": "^3.3.3",
+    "marked": "^13.0.3",
+    "react": "17.0.2",
+    "react-dom": "17.0.2",
+    "react-intl": "7.1.14"
```
xitij2000: Try to match these to the platform, so paragon 23+ and react ~18
Paragon 23+ brings design tokens which will make CSS handling easier.

frontend/.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
xitij2000: Any particular reason for this? I think it might be worth resolving the conflicts because this might cause issues later.

frontend/src/shortanswer/types.ts
xitij2000, line 1: I still see a lot of `?` which indicates an optional field. Since we control the server's output I feel these should be better defined. 

For instance for `ShortAnswerMessage` do we expect to see a message that has no source but only content? or content but no source? etc. If not then the `?` should be removed.

frontend/src/shortanswer/ShortAnswerStudioApp.tsx
```diff
@@ -0,0 +1,451 @@
...
+  if (!Array.isArray(value)) {
+    return [];
+  }
+
+  return value.map((entry) => {
+    return typeof entry === "string" ? entry : "";
+  });
+}
+
+function normalizeInitialState(initialState: ShortAnswerStudioState): ShortAnswerStudioState {
```
xitij2000: You could consider using [Yup](https://www.npmjs.com/package/yup) here. It's a data validation library that will really simplify this. Yup is used a lot in the Open edX ecosystem along with Formik, which you might also find useful for validating forms. 

With Yup you can do:

```typescript
const schema = Yup.object({
    display_name: Yup.string().ensure(),
    model: Yup.string().ensure(),
    model_api_key: Yup.string().ensure(),
    model_api_url: Yup.string().ensure(),
    question: Yup.string().ensure(),
    evaluation_prompt: Yup.string().ensure(),
    max_responses: Yup.string().ensure(),
    allow_reset: Yup.boolean(),
    character_image: Yup.string().ensure(),
    attachment_urls: Yup.array().of(Yup.string().ensure()),
});
```

Then you can apply `schema.validate(initialState)` and get back the cleaned state. 

With Formik you can apply this to forms. So if you have a field with validation: Yup.string().required("This field cannot be empty"), then it will automatically hook up the values, and validation and show that message when validation fails.

xitij2000: This could be used to validate in call places, the API response, the form and then the API post. 

However, I feel that the server response should already be validated by the server and if the form input is validated this way you don't need to validate more than once.

frontend/src/shortanswer/ShortAnswerStudioApp.tsx
```diff
@@ -0,0 +1,451 @@
...
+  const removeAttachmentLabel = intl.formatMessage({
+    id: "shortanswer.studio.removeAttachment",
+    defaultMessage: "Remove attachment",
+  });
+
+  return (
+    <Form.Group isInvalid={Boolean(errors && errors.length > 0)}>
+      <Form.Label>{metadata?.display_name || "Attachment URLs"}</Form.Label>
+      <div className="shortanswer-studio-attachment-list">
+        {attachmentUrls.map((attachmentUrl, index) => (
```
xitij2000, line 103: Formik FieldArray would be useful here: https://formik.org/docs/api/fieldarray

frontend/src/shortanswer/ShortAnswerStudentApp.tsx
```diff
@@ -0,0 +1,473 @@
...
+
+  return messages
+    .filter((message) => message && typeof message === "object")
+    .map((message) => {
+      const rawMessage = message as ShortAnswerMessage;
+      return {
+        source: rawMessage.source || "",
+        content: typeof rawMessage.content === "string" ? rawMessage.content : "",
+      };
+    });
```
xitij2000: Yup could be useful here. 

Also a cool trick I learned recently is to use flatMap to do map and filter at the same time. 

```suggestion
  return messages
    .flatMap((message) => {
       if (message && typeof message === "object") return [];
      const rawMessage = message as ShortAnswerMessage;
      return [{
        source: rawMessage.source || "",
        content: typeof rawMessage.content === "string" ? rawMessage.content : "",
      }];
    });
```

frontend/src/coding/CodingStudioApp.tsx
```diff
@@ -0,0 +1,365 @@
...
+          const normalizedResponse = normalizeStudioSaveResponse(error.payload);
+          setValidationErrors(normalizeValidationErrors(normalizedResponse.validation_errors));
+          setValidationWarnings(normalizeValidationWarnings(normalizedResponse.validation_warnings));
+          if (Object.keys(normalizedResponse.validation_errors).length > 0) {
+            inlineRequestError = intl.formatMessage({
+              id: "coding.studio.validationError",
+              defaultMessage: "Please fix the validation issues and try again.",
+            });
+          }
+        }
```
xitij2000: If you make this function async, you can use the cleaner async await API. 

```typescript
try { 
    const response = await submitStudioPayload(payload.handler_urls.studio_submit, buildSubmitPayload(values));
  // contents of then block
} catch(error) {
  // contents of catch block
}
```

frontend/src/coding/CodingStudioApp.tsx
xitij2000, line 1: This is a great candidate for Yup, and Formik.

xitij2000 (commented):
I think there is a lot of data validation here which makes using something like Formik and Yup worthwhile. 

A lot of types just `?` for every field. I think this needs to be tightened, otherwise we lose a lot of the benefits of typescript.

.nvmrc
pkulkark: We need to maintain compatibility with redwood. We should probably create a new ulmo-release branch as a follow up and update the versions there.

frontend/package.json
pkulkark: Same thing here - chose the versions compatible with redwood. We can create an ulmo-branch and update the versions there in a follow-up.

frontend/.npmrc
pkulkark: Sorry that slipped in during testing somehow. Removed it.

frontend/src/shortanswer/types.ts
pkulkark: I've reduced it wherever possible.

frontend/src/shortanswer/ShortAnswerStudioApp.tsx
pkulkark: Good idea. I wasn't aware of Yup and Formik. I refactored the code to use them.

frontend/src/coding/CodingStudioApp.tsx
pkulkark: Good idea. Updated.

frontend/src/coding/CodingStudioApp.tsx
pkulkark: Thanks for the suggestion. I've refactored it to use Yup and Formik.

.nvmrc
xitij2000: Since this isn't using anything from the platform or interacting with it, you can use a different version. For example, (open)edx-platform is still using Paragon 2 while MFEs are on Paragon 23+. XBlocks are isolated so you can pick a newer version. You can pick any version here since this will only be used to compile the components and only the compiled static code will be used in the XBlock.

frontend/package.json
xitij2000: You don't need to worry about using the same versions since this is an isolated component.

ai_eval/static/css/coach_studio.css
xitij2000: What I meant here is that you can set these to paragon variables by default and use these colours via the client's theme. 

For example, if you do the following:

```css
--pgn-color-coaching-border: var(--pgn-color-gray-100);
```

Then the default colour of the block can be taken from Paragon, which will ensure that they fit will with most instances. However, any client can then supply their own custom value for ``--pgn-color-coaching-border`` and theme it to their spec. 

Since this is currently not using the system theme, you can give these values as fallback:

```css
--pgn-color-coaching-border: var(--pgn-color-gray-100, #cbcbcb);
```

Prefixing with --pgn means that you can configure it with design tokens since all variables produced from  tokens by Paragon are prefixed with pgn. Setting it to the default gray color means that if configured this sill pick up the client theme and giving it the fallback value means that you still get exactly the same UI for now.

ai_eval/base.py
xitij2000: What I meant was, instead of having a `_handler_url` method, you can replace calls to `self._handler_url` with a direct call to `self.runtime.handler_url`

ai_eval/base.py
```diff
@@ -264,92 +283,265 @@ def _get_studio_lock_payload(self) -> dict:
...
+        """
+        Build a handler URL for view payloads.
+        """
+        return self.runtime.handler_url(self, handler_name)
+
+    def _static_url(self, path: str) -> str:
+        """
+        Return a cacheable URL for a static resource.
+        """
+        return self.runtime.local_resource_url(self, path)
```
xitij2000: Same here. You can just call `self.runtime.local_resource_url` directly.

ai_eval/coach.py
```diff
@@ -467,44 +610,157 @@ def _get_character_data(self, character_index):  # pylint: disable=missing-funct
...
+            attempts_used = int(attempts_used or 0)
+        except (TypeError, ValueError):
+            attempts_used = 0
+        normalized["attempts_used"] = max(attempts_used, 0)
+        normalized["finished"] = bool(session.get("finished", False))
+        normalized["final_submission"] = str(session.get("final_submission") or "")
+        normalized["final_evaluation_markdown"] = str(
+            session.get("final_evaluation_markdown") or ""
+        )
+        return normalized
```
xitij2000: You can use Pydantic for this:

```
class Session(BaseModel):
    workspace_history: list[str] = []
    coach_history: list[str] = []
    evaluation_fragments: list[str] = []
    attempts_used: Annotated[int, Field(ge=0)] = 0
    finished: bool = False
    final_submission: str = ""
    final_evaluation_markdown: str = ""
```

Adding some validators will let you get the list and int conversion

ai_eval/coach.py
```diff
@@ -467,44 +610,157 @@ def _get_character_data(self, character_index):  # pylint: disable=missing-funct
...
+        """Ensure Coaching learner state has a normalized active session."""
+        current_sessions = self.sessions if isinstance(self.sessions, list) else []
+        normalized_sessions = []
+        sessions_changed = not isinstance(self.sessions, list)
+
+        for session in current_sessions:
+            normalized_session = self._normalize_session(session)
+            normalized_sessions.append(normalized_session)
+            if normalized_session != session:
+                sessions_changed = True
```
xitij2000: If you have a Pydantic model for sessions you can then apply it in bulk here with a [TypeAdapter](https://pydantic.dev/docs/validation/latest/concepts/type_adapter/)

ai_eval/coach.py
```diff
@@ -467,44 +610,157 @@ def _get_character_data(self, character_index):  # pylint: disable=missing-funct
...
+        """Append and return a fresh active learner session."""
+        sessions = list(self.sessions or [])
+        sessions.append(self._empty_session())
+        self.sessions = sessions
+        return sessions[-1]
+
+    def _session_has_meaningful_data(self, session):
+        """Return whether a session contains learner progress worth preserving."""
+        session = self._normalize_session(session)
+        return any([
```
xitij2000: I think you can just compare it to an empty session, i.e. `return session != self._empty_session()`

ai_eval/coach.py
```diff
@@ -467,44 +610,157 @@ def _get_character_data(self, character_index):  # pylint: disable=missing-funct
...
+
+        if migrate_from_older_state:
+            self._clear_older_state_runtime_fields()
+            self.save()
+
+    def _get_active_session(self):
+        """Return the normalized active learner session."""
+        self._ensure_sessions_initialized()
+        sessions = list(self.sessions or [])
+        active_session = self._normalize_session(sessions[-1])
```
xitij2000: This seems to be used a lot. Can we just ensure that the data is clean when being saved to self.sessions so we can assume it's ready to use everywhere else?

ai_eval/coach.py
```diff
@@ -694,59 +952,146 @@ def _clear_thread_contexts(self, contexts):
...
+        if field_name == "scenario_data" and not isinstance(parsed_value, dict):
+            return None, _("Scenario data must be a JSON object (dictionary).")
+
+        if field_name == "blacklist" and not isinstance(parsed_value, list):
+            return None, _("Output blacklist must be a JSON array.")
+
+        if field_name == "blacklist":
+            parsed_value = self._sanitize_blacklist_values(parsed_value)
+
+        return parsed_value, None
```
xitij2000, line 1015: Pydantic would be good here.

frontend/src/coaching/CoachingStudentApp.tsx
```diff
@@ -0,0 +1,1176 @@
...
+    );
+  }
+
+  return <div className="coach-message__avatar coach-message__avatar--empty" aria-hidden="true" />;
+}
+
+const ChatMessage = React.memo(function ChatMessage({ message }: { message: PendingCoachingMessage }) {
+  const pane = message.pane === "coach" ? "coach" : "workspace";
+  const isUser = Boolean(message.is_user);
+  const character = normalizeCharacter(message.character, pane);
```
xitij2000: I think you should clean up the data once during input and then the parameters that React components get should already be in a clean state.

frontend/src/shortanswer/types.ts
```diff
@@ -0,0 +1,63 @@
+import { SharedPayload, StudioFieldMetadata } from "../shared/types";
+
+export interface ShortAnswerMessage {
+  source: string;
+  content: string;
+}
+
+export interface ShortAnswerStudentHandlerUrls {
+  get_response?: string;
+  reset?: string;
```
xitij2000: Why would these be undefined?

frontend/src/shortanswer/types.ts
```diff
@@ -0,0 +1,63 @@
...
+  evaluation_prompt: string;
+  max_responses: number | string | null;
+  model: string;
+  model_api_key: string;
+  model_api_url: string;
+  question: string;
+}
+
+export interface ShortAnswerStudioHandlerUrls {
+  studio_submit?: string;
```
xitij2000: Likewise here, this shouldn't be undefined.

frontend/src/shortanswer/studio.tsx
```diff
@@ -0,0 +1,39 @@
...
+function normalizePayload(
+  runtime: XBlockRuntime,
+  element: XBlockElementLike,
+  data: unknown,
+): ShortAnswerStudioPayload {
+  const payloadData = (data || {}) as StudioPayloadInput;
+
+  return {
+    view: payloadData.view || "studio",
+    handler_urls: payloadData.handler_urls || {
```
xitij2000: Why not just use `runtime.handlerUrl` and skip sending it from the server?

frontend/src/shortanswer/student.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+function normalizePayload(
+  runtime: XBlockRuntime,
+  element: XBlockElementLike,
+  data: unknown,
+): ShortAnswerStudentPayload {
+  const payloadData = (data || {}) as ShortAnswerPayloadInput;
+
+  return {
+    view: payloadData.view || "student",
```
xitij2000: Since we already know this is the student view, can't we just skip this and set it directly to "student"?

frontend/src/shortanswer/student.tsx
```diff
@@ -0,0 +1,50 @@
...
+    handler_urls: payloadData.handler_urls || {
+      get_response: runtime.handlerUrl(element, "get_response"),
+      reset: runtime.handlerUrl(element, "reset"),
+    },
+    initial_state: payloadData.initial_state || {
+      messages: payloadData.messages as ShortAnswerMessage[] | undefined,
```
xitij2000: Since you're intentionally setting a value here, it's better to use null instead of undefined.

frontend/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "name": "xblock-ai-eval-frontend",
+  "private": true,
+  "scripts": {
+    "build": "webpack --mode production",
+    "build:dev": "webpack --mode development",
+    "typecheck": "tsc --noEmit",
+    "watch": "webpack --mode development --watch"
+  },
+  "dependencies": {
+    "@openedx/paragon": "^22.8.1",
```
xitij2000, line 13: You can use Paragon 23+. Since this is an isolated component you don't need to worry about compatibility.

frontend/src/shared/renderMarkdown.ts
```diff
@@ -0,0 +1,11 @@
+import DOMPurify from "dompurify";
+import { marked } from "marked";
+
+export function renderMarkdown(markdown?: string): string {
```
xitij2000: Why would markdown be unknown or even null here?

frontend/src/shared/renderMarkdown.ts
```diff
@@ -0,0 +1,11 @@
...
+  if (!markdown) {
+    return "";
+  }
+
+  const html = marked.parse(markdown, { async: false }) as string;
```
xitij2000, line 9: Why not use async here?

frontend/src/shared/request.ts
```diff
@@ -0,0 +1,98 @@
...
+      return {};
+    }
+
+    try {
+      return JSON.parse(text) as UnknownRecord;
+    } catch (error) {
+      return { raw: text };
+    }
+  });
+}
```
xitij2000: Prefer async/await instead of using promise.then.catch

frontend/src/shared/request.ts
```diff
@@ -0,0 +1,98 @@
...
+    return response.statusText;
+  }
+
+  return "Request failed.";
+}
+
+function getCookie(name: string) {
+  if (typeof document === "undefined" || !document.cookie) {
+    return null;
+  }
```
xitij2000, line 55: Why would document be undefined? Is it for testing? If so you can mock it.

frontend/src/shared/request.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  if (csrfToken) {
+    headers["X-CSRFToken"] = csrfToken;
+  }
+
+  return fetch(url, {
+    method: "POST",
+    credentials: "same-origin",
+    headers,
+    body: JSON.stringify(payload || {}),
```
xitij2000: Instead of doing this, give payload a default value of {}.

frontend/src/shared/request.ts
```diff
@@ -0,0 +1,98 @@
...
+    const cookie = cookies[index].trim();
+    if (cookie.startsWith(prefix)) {
+      return decodeURIComponent(cookie.slice(prefix.length));
+    }
+  }
+
+  return null;
+}
+
+export function postJson<TResponse extends UnknownRecord = UnknownRecord>(
```
xitij2000: Function can be made async.

frontend/src/shared/studio.ts
```diff
@@ -0,0 +1,144 @@
...
+export function getSaveErrorMessage(error: unknown, fallbackMessage: string) {
+  if (error instanceof RequestError && error.payload) {
+    if (hasStudioValidationIssues(error.payload)) {
+      return error.message;
+    }
+  }
+
+  if (error instanceof RequestError && error.message) {
+    return error.message;
+  }
```
xitij2000: Where is this function used? I see it redefined in multiple places.

frontend/src/shared/useStudioModalActions.ts
```diff
@@ -0,0 +1,71 @@
...
+}: UseStudioModalActionsOptions) {
+  useEffect(() => {
+    const root = document.querySelector(rootSelector);
+    const modal = root?.closest(".edit-xblock-modal");
+    const modalActions = modal?.querySelector(".modal-actions") as HTMLElement | null;
+    const saveAction = modalActions?.querySelector(".action-save") as HTMLAnchorElement | null;
+    const cancelAction = modalActions?.querySelector(".action-cancel") as HTMLAnchorElement | null;
+
+    if (!modalActions || !saveAction || !cancelAction) {
+      return undefined;
```
xitij2000: Avoid undefined, prefer null.

frontend/src/coding/CodingStudentApp.tsx
```diff
@@ -0,0 +1,796 @@
...
+        break;
+      case "End":
+        nextIndex = tabOrder.length - 1;
+        break;
+      default:
+        return;
+    }
+
+    onActivateTab(tabOrder[nextIndex], true);
+  }
```
xitij2000: Consider using https://paragon-openedx.netlify.app/components/hooks/usearrowkeynavigation/

frontend/src/coding/CodingStudentApp.tsx
```diff
@@ -0,0 +1,796 @@
...
+              event.key === "ArrowLeft" ||
+              event.key === "ArrowUp" ||
+              event.key === "Home" ||
+              event.key === "End"
+            ) {
+              event.preventDefault();
+            }
+            moveFocus("output", event.key);
+          }}
+        >
```
xitij2000, line 277: Avoid arrow functions in react code in favour of [useCallback](https://react.dev/reference/react/useCallback)

frontend/src/coding/CodingStudentApp.tsx
```diff
@@ -0,0 +1,796 @@
...
+        },
+      );
+      const statusId = Number(result.status?.id || 0);
+
+      if (statusId === 1 || statusId === 2) {
+        if (retries >= MAX_JUDGE0_RETRY_ITER) {
+          throw new Error(
+            "Judge0 submission result fetch failed after " +
+              String(MAX_JUDGE0_RETRY_ITER) +
+              " attempts.",
```
xitij2000: You should use a template string here.

frontend/src/coding/CodingStudentApp.tsx
xitij2000, line 1: These components are doing too much, split it into one file per component ideally or a few very small related components. 

Also please move out the web interactions into a separate file.

xitij2000 (commented):
I don't see any tests for the frontend code. Could you please add some?
I still see a lot of data validation cleanup etc. inside components which isn't ideal. 

I think the best way to solve this is to have a clear boundary between the views and the API interactions. 

When you get data from the XBlock init function it should already be in the correct format, so you can simply set the correct types there. 

Since we're not using React Query here, I think you can divide the code as follows:
- Functions that purely do API requests to endpoints like studio_submit 
- Functions that clean up the data
- Functions that perform the connect the above two by cleaning up data and return cleaned data
- Hooks that use the above and track requests. 

For example:
```typescript
async function postStudioSubmit(data: StudioSubmitPayload) {
  const response = await fetch("URL");
  return await response.json();
}

function cleanupData(rawData: any) {
  // process data
  return cleanData as StudioResponse;
}

async function studioSubmit(data: StudioSubmitPayload) {
    const response = await postStudioSubmit(data);
    return cleanupData(response);
}

function useStudioSubmit() {
  const [error, setError] = useState<string | null>(null);
  const [data, setData] = useState<StudioResponse | null>(null);
  const [isLoading, setIsLoading] = useState(false);
  const submitToStudio = async (data: StudioSubmitPayload) => {
    setIsLoading(true);
    setError(null);
    try {
      const data = await studioSubmit(data);
      setData(data);
    } catch (err) {
      setError(err);
    }
    setIsLoading(false);
  };
  return {
    error,
    data,
    isLoading,
    submitToStudio,
  };
}
```

If you want you can move the try catch one step up and return clean errors to the hook. 
You can also generalise this hook by calling it `useRequest(requestFn)` and have `const data = await requestFn()` etc.

ai_eval/base.py
pkulkark: Ah ok. Removed the wrapper and moved the call inline.

ai_eval/static/css/coach_studio.css
pkulkark: Done. Prefixed all variables with --pgn.

ai_eval/base.py
pkulkark: Done.

ai_eval/coach.py
pkulkark: I've added Pydantic validators.

ai_eval/coach.py
pkulkark: I went with `model_validate` so that valid sessions are preserved if one is corrupt.

ai_eval/coach.py
pkulkark: Done.

ai_eval/coach.py
pkulkark: Done.

ai_eval/coach.py
pkulkark: Done.

frontend/src/coaching/CoachingStudentApp.tsx
pkulkark: Removed them. Backend sends complete, typed data; frontend trusts it.

frontend/src/shortanswer/types.ts
pkulkark: removed `?`

frontend/src/shortanswer/types.ts
pkulkark: removed `?`

frontend/src/shortanswer/studio.tsx
pkulkark: So that the backend remains the single source of truth.

frontend/src/shortanswer/student.tsx
pkulkark: Done.

frontend/src/shortanswer/student.tsx
pkulkark: Done

frontend/src/shared/renderMarkdown.ts
pkulkark: There was one call (evaluation markdown) that passed a potentially undefined value but I fixed it in the call site and removed `?` from here.

frontend/src/shared/renderMarkdown.ts
pkulkark: Making it async would require useEffect + state at every call site. The markdown payloads are simple chat messages and evaluation text.

frontend/src/shared/request.ts
pkulkark: Updated it.

frontend/src/shared/request.ts
pkulkark: Removed it.

frontend/src/shared/request.ts
pkulkark: Done.

frontend/src/shared/request.ts
pkulkark: Done.

frontend/src/shared/studio.ts
pkulkark: Consolidated the duplicate helpers into one.

frontend/src/shared/useStudioModalActions.ts
pkulkark: Returning null caused an error. Apparently react expects `void | (() => void)` for useEffect return.

frontend/src/coding/CodingStudentApp.tsx
pkulkark: Done. Replaced with Paragon's useArrowKeyNavigation.

frontend/src/coding/CodingStudentApp.tsx
pkulkark: The only React.memo component (ChatMessage) receives no callbacks.

frontend/src/coding/CodingStudentApp.tsx
pkulkark: Extracted API request functions into separate files. Kept small presentational components co-located with their parent.

ai_eval/coach.py
```diff
@@ -78,10 +79,58 @@ class EvaluationCriterion(pydantic.BaseModel):
...
+        return data
+
+
+class CoachingSession(pydantic.BaseModel):
+    """Normalized shape of a single Coaching learner session."""
+
+    workspace_history: list = pydantic.Field(default_factory=list)
+    coach_history: list = pydantic.Field(default_factory=list)
+    evaluation_fragments: list = pydantic.Field(default_factory=list)
+    attempts_used: int = pydantic.Field(default=0, ge=0)
```
samuelallan72, line 105: @pkulkark cc @xitij2000 if we're going to add Pydantic, please model all the fields completely. This only goes halfway, because it only checks the top level structure.

Also, please be stricter with what data is accepted. Pydantic as used here is not that useful, because there are all the extra custom field validators that are quite lenient. I don't think these are necessary, and they add more verbosity.

ai_eval/coach.py
samuelallan72: @pkulkark cc @xitij2000 By using pydantic here, we should be able to do something like this in the `studio_submit` handler:

```python
try:
  validated_data = MyPydanticDataModel(**data)
except validationerrors as e
  errors = e.errors()
  # process and return errors for the frontend

# set the fields, etc.
self.XYZ = validated_data.XYZ
```

And all the custom validating as above in `_studio_parse_json_field` would be either handled by pydantic or in custom validators.

This is a large refactoring though, and probably better in a follow up PR(?).

ai_eval/coach.py
xitij2000: @samuelallan72 The validation was moved from logic elsewhere to Pydantic since pydantic is already being used. I agree that Pydantic should be used more aggressively for validation, and we should add a task to fully implement it since the scope of this one is already too large.

ai_eval/coach.py
xitij2000: Definitely would love to see the validation logic all moved to Pydantic, but as you said, it's large enough already so we should do that in a follow up.

frontend/src/coaching/coachingStudioAdapters.ts
```diff
@@ -0,0 +1,175 @@
...
+
+      return createInvalidListItem(value, "Saved criterion must include a name.");
+    }
+
+    return createInvalidListItem(value, "Saved criterion must be an object with a name.");
+  });
+}
+
+export function getScenarioEditorModel(
+  scenarioData: ScenarioData | undefined,
```
xitij2000:
```suggestion
  scenarioData?: ScenarioData,
```

frontend/src/coaching/coachingStudioAdapters.ts
```diff
@@ -0,0 +1,175 @@
...
+): CoachingScenarioEditorModel {
+  return {
+    caseDetails: scenarioData?.case_details ?? "",
+  };
+}
+
+export function getScenarioListItems(
+  scenarioData: ScenarioData | undefined,
```
xitij2000, line 98:
```suggestion
  scenarioData?: ScenarioData,
```

frontend/src/coaching/CoachingStudioApp.tsx
```diff
@@ -0,0 +1,1516 @@
...
+      return true;
+    }),
+  workspace_title: Yup.string().ensure(),
+});
+
+function updateListItemAtIndex(items: CoachingListItem[], index: number, nextValue: string) {
+  return items.map((item, itemIndex) => {
+    return itemIndex === index ? { value: nextValue } : item;
+  });
+}
```
xitij2000, line 100: Do we need immutability here? Of not we can just simplify this to:
``items[index] = nextValue``

frontend/src/coaching/CoachingStudioApp.tsx
```diff
@@ -0,0 +1,1516 @@
...
+
+function appendListItem(items: CoachingListItem[]) {
+  return [...items, { value: "" }];
+}
```
xitij2000, line 104: Likewise this can be `item.push({value: ""})`

frontend/src/coaching/CoachingStudioApp.tsx
```diff
@@ -0,0 +1,1516 @@
...
+
+function removeListItemAtIndex(items: CoachingListItem[], index: number) {
+  return items.filter((_, itemIndex) => {
+    return itemIndex !== index;
+  });
```
xitij2000, line 109:
```suggestion
  return items.toSpliced(index, 1);
```

frontend/src/coaching/CoachingStudioApp.tsx
```diff
@@ -0,0 +1,1516 @@
...
+
+  return (
+    <Form.Group className="coaching-studio-field" isInvalid={Boolean(errors?.length)}>
+      <div className="coaching-studio-field-header">
+        <div className="coaching-studio-field-title-row">
+          <Form.Label className="coaching-studio-field-label">{label}</Form.Label>
+        </div>
+        <FieldDescription description={description} helpText={metadata?.help} />
+      </div>
+      <div className="coaching-studio-field-control">
```
xitij2000, line 422: There is a lot of common code here, this can be DRY'd up. 
Noted for future PR.

frontend/src/coding/CodingStudioApp.tsx
```diff
@@ -0,0 +1,403 @@
...
+  return {
+    display_name: values.display_name || "",
+    model: values.model || "",
+    model_api_key: values.model_api_key || "",
+    model_api_url: values.model_api_url || "",
+    question: values.question || "",
+    evaluation_prompt: values.evaluation_prompt || "",
+    judge0_api_key: values.judge0_api_key || "",
+    language: values.language || "",
+  };
```
xitij2000: You can use Yup here. I think `codingSchema.case(values)` should work.

frontend/src/shared/renderMarkdown.ts
xitij2000: Sure, but you can simplify it with a custom hook:

```typescript
function useRenderedMarkdown(markdown: string) {
  const [html, setHtml] = useState('');
  useEffect(() => {
    const render = async () => {
        setHtml(DOMPurify.sanitize(await marked.parse(markdown??"", { async: true }) as string;))
    };
    render(); 
  }, [text]);
  return html;
}
```

Then you can just use ``const html = useRenderedMarkdown(markdown)``. 

In any case since this will work, we can leave that for the future.

frontend/src/shared/useStudioModalActions.ts
xitij2000: Then you can simply use "return" undefined is the default return value.

xitij2000 (approved):
I think there is still some refactoring that would be helpful here. It is a complex codebase with a lot of scope for extracting higher-order components.

A lot of files contain too many components. We can break this down to make the code easier to handle. 

A well-defined schema in the backend with well-defined types in the frontend would simplify a lot of the code that is simply validating and reformatting data. A lot of fields are optional, which makes me feel like there is scope for restructuring.

 For instance StudioLockMetadata has all fields optional, and it itself is an optional field of ShortAnswerStudioMeta. This means there is a chain of undefined. undefined could come from a missing api response, or a missing return value etc, null is more explicit and means something was deliberately left out. 

I still need to test these latest changes on the server. However from a cost perspective we can always continue refactoring and improving over time and future PRs. 

Thanks for bearing with me! It's taken some time since this is a very complex UI and I needed time to wrap my head around some thing, and I'm still doing that.

frontend/src/shortanswer/student.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+type ShortAnswerPayloadInput = Partial<ShortAnswerStudentPayload> & {
+  allow_reset?: boolean;
+  character_image?: string;
+  max_responses?: number;
+  messages?: unknown[];
+  question?: string;
+};
+
+function normalizePayload(
```
samuelallan72: @pkulkark drive-by comment, since I came across this working on #25. Why is there so much normalising happening across the app? For example, here this should not be required at all, since this is data that we completely control, provided directly by the backend.

Why not build the full data in the backend, and the corresponding type in the frontend, then there is no need for all this double processing.

All the indirection, normalising, etc. makes it difficult to neatly extend the code, add new fields etc. :/

cc @xitij2000

ai_eval/base.py
```diff
@@ -252,7 +269,9 @@ def _get_model_key_presence_map(self) -> dict[str, bool]:
             for model in models
         }
 
-    def _get_studio_lock_payload(self) -> dict:
+    # Studio metadata helpers
+
+    def _studio_lock_metadata(self) -> dict:
```
samuelallan72, line 274: @pkulkark what exactly is lock metadata in this context?

frontend/src/shortanswer/student.tsx
```diff
@@ -0,0 +1,50 @@
...
+import ShortAnswerStudentApp from "./ShortAnswerStudentApp";
+import { ShortAnswerMessage, ShortAnswerStudentPayload } from "./types";
+
+type ShortAnswerPayloadInput = Partial<ShortAnswerStudentPayload> & {
+  allow_reset?: boolean;
+  character_image?: string;
+  max_responses?: number;
+  messages?: unknown[];
+  question?: string;
+};
```
samuelallan72: @pkulkark could you explain how this type works? Why are we redefining fields from ShortAnswerStudentPayload.meta inline here, then using these as fallback values if the .meta fields aren't set? What sets these?

samuelallan72:
@pkulkark @xitij2000 another drive-by comment: it would be helpful to add a code formatter, frontend linter, and CI for the frontend (typecheck, check formatting, build).

xitij2000:
> @pkulkark @xitij2000 another drive-by comment: it would be helpful to add a code formatter, frontend linter, and CI for the frontend (typecheck, check formatting, build).

I think the [exemplar blocks](https://gitlab.com/opencraft/dev/exemplar-blocks) have a pretty good structure for this, it would be good to follow the same template here, since it includes quality checks testing etc. However it does use vite including for testing while this uses webpack so might be more complex than expected. I like that it will automatically generate frontend types from backend types.

frontend/src/coaching/coachingStudioAdapters.ts
pkulkark: Couldn't apply this because it has a required `fieldName` parameter after. TypeScript does not allow a required parameter after an optional one.

frontend/src/coaching/CoachingStudioApp.tsx
pkulkark: This is to ensure Formik doesn't skip re-rendering as it uses reference comparison to decide whether to re-render. Mutating the original array in-place could result in Formik seeing the same reference and skipping re-rendering.

frontend/src/coaching/CoachingStudioApp.tsx
pkulkark: Same Formik immutability concern here.

frontend/src/coaching/CoachingStudioApp.tsx
pkulkark: We would need `ES2023` for this. Also risks crashing in slightly older browsers. But the `.filter()` approach is universally compatible.

frontend/src/coding/CodingStudioApp.tsx
pkulkark: Oh cool. Updated it to use Yup.

frontend/src/shortanswer/student.tsx
pkulkark: Thanks for the review. A bit of context on this work - this PR covers both a full jQuery-to-React frontend migration and a redesign of the authoring flow. Because the client has this running in production, I wanted to ensure existing stored XBlock data wouldn't break during the transition. I approached it in stages. The normalizers were added during one of these stages. The backend was tightened in later commits/review rounds (Pydantic validation, complete payloads with all required fields, etc) and this got buried.
In hindsight, this should have been split across multiple tasks/PRs - I didn't expect the scope to grow this much. Anyway I have now removed these normalizers for payload. The backend sends the complete typed payload, the frontend trusts it.

frontend/src/shortanswer/student.tsx
pkulkark: Removed in the last commit.

ai_eval/base.py
pkulkark: It tells the Studio frontend which form fields to disable (lock). This is needed for API key fields - when a site admin configures a global API key via Django settings or site configuration, course authors shouldn't be able to override it per-block, so the field is rendered as disabled in the editor.

pkulkark:
@xitij2000 I appreciate all the review comments. I was trying to learn while building it out and also knock out multiple requirements (react frontend migration + redesign) at the same time which turned this into a mammoth PR - I should've split them into smaller manageable tasks. Anyway I think this is stable enough to merge for now. I'll create a follow up task with the remaining improvements noted.

frontend/src/coaching/coachingStudioAdapters.ts
xitij2000: You can just reorder them?

PR #25, opened 2026-04-20
Add support for downloading pdf transcripts

samuelallan72:
## Description

Add support for downloading pdf transcripts of the student interactions
with the AI Evaluation XBlocks.

PDF downloads can be enabled/disabled per-component by instructors,
and there is a customisable title and description for the download pdf
section for students.

The PDFs support a configurable logo image in the header.

TODO before merge:

- [x] add tests
- [x] I can add some screenshots to the PR description if desired too - that may be helpful for future reference
- [x] deploy to a sandbox for testing by product
- [x] resolve the TODO/XXX comments in the code (some may require follow ups; as long as there's a plan for each)

<details>
<summary>Screenshots</summary>

image

---

image

---

image

</details>

## Supporting information

Private-ref: https://tasks.opencraft.com/browse/BB-10694

Private-ref: [UI designs in Figma](https://www.figma.com/design/J4D6OOPXu7UXd9NxwPaRBd/WGU-XBlocks-2025-2026?node-id=2080-320&t=hSKZ303xou3w6VWt-1)

## Testing instructions

- install this xblock in an openedx sandbox / devstack
- add `["shortanswer_ai_eval",  "coding_ai_eval", "coach_ai_eval"]` to the advanced module list in a course in authoring
- add an instance of each AI evaluation blocks (coached, coding, short answer) to a course in studio
- for each, configure them with valid information, including an llm service api key, etc.
- save them; now we're ready to test the pdf functionality.
- for each component: play with the pdf related controls in the studio editor (pdf allowed, section title, section description). Save and verify the changes persist.
- For the short answer and coding blocks: verify the pdf section displays when pdfs are allowed, and hidden when not allowed.
  - For the coached block, the pdf section should be only displayed if pdf downloads are allowed AND the student has submitted the assignment
- View the pdfs and verify they match the designs (see supporting information). Note that the designs have changed slightly over time (the originals had very small font).
- Try to break the pdfs: enter malformed and malicious markdown/html in the chat boxes, try generating the pdfs from empty states, enter long content (in chat boxes, code boxes, display names, etc.), use emoji, etc. etc.
- Test all the links in the PDFs (including all the breadcrumb navigation links)
- Follow the new instructions in the readme for installing custom fonts to have full support for emojis and the desired font (Inter). Regenerate some pdfs and verify emojis display correctly and the correct font is used (at least, the font should be now different to previous pdfs generated).
- Follow the new instructions in the readme for setting a custom pdf header logo. Regenerate some pdfs and verify the header logo updates (note that there seems to be aggressive caching for the site configuration; you may need to wait some time or restart your devstack).
- Check the code is documented enough
- check the formatting of the updated readme

## Other notes

- There are several TODO and XXX comments inline. These are some minor things that I haven't got to yet or have questions about - hoping to discuss during the review. Some may be larger scope and worth a follow up PR.


ai_eval/coach.py
```diff
@@ -1237,3 +1240,46 @@ def _evaluator_messages():
...
+        """Generate and download the pdf summary of the exercise."""
+
+        # TODO: check if we can download here; we only want to enable download if the user has submitted and not started a new session
+
+        # https://openedx.atlassian.net/wiki/spaces/PLAT/pages/113607155/How+do+I+access+student+data+from+within+an+XBlock
+        user = self.runtime.service(self, "user").get_current_user()
+        user_email = user.emails[0] if user.emails else ""
+
+        # Location information
+        unit = self.get_parent()
```
Kelketek: What happens if this block is being tested in the workbench, or in a library?

ai_eval/coach.py
```diff
@@ -1237,3 +1240,46 @@ def _evaluator_messages():
...
+
+        metadata = Metadata(
+            info=Info(
+                title=self.display_name,
+                author="WGU",
+                description="PDF transcript of a student submission to a Coached AI XBlock",
+            ),
+            student=Student(email=user_email),
+            branding=Branding(
+                logo_image_src="https://www.wgu.edu/etc.clientlibs/wgu/clientlibs/clientlib-site/resources/images/wgu-logo.svg"
```
Kelketek: The default image should probably be a setting, with a different generic value by default, so that anyone installing this block doesn't get the WGU image.

ai_eval/coach.py
```diff
@@ -1237,3 +1240,46 @@ def _evaluator_messages():
...
+                subsection_name=subsection.display_name,
+                subsection_url="https://example.com",
+            ),
+        )
+        content = CoachContent()
+        pdf_data: bytes = generate_coach_pdf(metadata, content)
+
+        __import__("pprint").pprint(course)
+
+        # TODO: generate a dynamic but safe filename, based on the display name and student name?
```
Kelketek: The slugify Django utility is a good candidate for this.

ai_eval/pdf_generator.py
```diff
@@ -0,0 +1,143 @@
...
+    location: Location
+
+
+# TODO: we'll need to use something like this when processing the actual chat messages.
+# The new frontend is tentatively using dompurify and marked https://github.com/open-craft/xblock-ai-evaluation/blob/27bc537ad06c1d1d62dc1d68d86a43cb79daff4a/frontend/src/shared/renderMarkdown.ts#L1
+# (we may wish to consolidate markdown rendering and html cleaning to the backend, and remove it from the frontend).
+def _markdown_to_safe_html(markdown_content: str) -> str:
+    html_content = mistune.html(markdown_content)
+    cleaner = nh3.Cleaner()
+    # TODO: ensure we only allow a safe subset of html here. Don't allow images, styles, any attributes - all we need is enough to cover basic html structure generated from markdown.
```
Kelketek: This says 'TODO' but isn't nh3 already cleaning it like that?

ai_eval/pdf_generator.py
```diff
@@ -0,0 +1,143 @@
...
+    now = datetime.now(UTC)
+    date_info = _Date(
+        machine=now.isoformat(), pretty=now.strftime("%d %B %Y, %I:%M%p %Z")
+    )
+
+    html = _JINJA_ENV.get_template(content.template_file()).render(
+        content=content.model_dump(), date=date_info, **metadata.model_dump()
+    )
+
+    # XXX: save html for debugging; remove before merging
```
Kelketek: Why `XXX` instead of `TODO`? `TODO` is recognized by some linters to prevent merging without addressing the listed items, though I don't know that we're using it that way here.

Kelketek (commented):
Hi @samuelallan72 , just had some questions/comments on the WIP.

ai_eval/coach.py
samuelallan72: This part will fail... I can add some fallback conditions. I'm not sure if there's a better way to get the course/section/subsection information here?

ai_eval/coach.py
samuelallan72: Ah nice, thanks!

ai_eval/pdf_generator.py
samuelallan72: Mostly - I don't remember exactly the defaults - by default it would definitely be _safe_ html. But I was thinking we might want to only allow a smaller subset of html to avoid the layout getting messed up, or pulling in unwanted images, etc.

ai_eval/pdf_generator.py
samuelallan72: No strong reason. XXX is one of the common tags, along with TODO, FIXME, and maybe others, that these kind of linters should pick up. I think I went for XXX because it's not something new that needs doing, it's more flagging code that needs fixing I guess.

samuelallan72: Ah yes, these are configured in the pylintrc: https://github.com/open-craft/xblock-ai-evaluation/blob/2064588b55facfb4634fb99803575d4468669dbb/pylintrc#L322-L323

ai_eval/pdf_generator.py
samuelallan72: @pkulkark do you know the requirements and intentions for html/markdown support here? I'm guessing we don't want to support images, custom css, etc. Is markdown just to support providing headings, lists, and such?

As in, can we strip out anything that wouldn't be normal base markdown formatting (lists, headings, bold, italic, etc.)?

If we don't do very strict stripping, we need to implement tighter controls around pdf generation - whitelist/blacklist urls, timeouts, etc. Otherwise it would be possible to do things like malicious src attributes to leak internal files.

ai_eval/coach.py
Kelketek: No idea. But fallback is a good idea.

ai_eval/pdf_generator.py
samuelallan72: I did some digging here: nh3 does strip out most of what I was worried about. Additionally, I stripped `<a>` and `` tags, because I don't think they're really going to be used, and they are easy surfaces for trying various attacks.

ai_eval/coach.py
samuelallan72: Updated :)

ai_eval/coach.py
samuelallan72: Added a fallback :)

ai_eval/coach.py
```diff
@@ -1237,3 +1245,190 @@ def _evaluator_messages():
...
+            or "Student"
+        )
+
+        # Follow a simplified method of ordering the chats if timestamps are not available:
+        # simply show the coach chat first, then show the workspace chat (no interleaving by time order).
+        # TODO: is this necessary? If this block is not being used in production yet,
+        #       we can simply make the breaking changes, and let any test users know they need to reset the components.
+        no_timestamps = (
+            session["workspace_history"] and not session["workspace_history"][0].get("time")
+        ) or (session["coach_history"] and not session["coach_history"][0].get("time"))
```
samuelallan72: @pkulkark what's the status of this block deployments-wise? Do we need to worry about backwards compatibility for existing problems?

If this is still early days testing, perhaps we can take the opportunity to simplify a lot of logic while making a batch of breaking changes (eg. this kind of conditionals to support older submissions with missing values, the normalising code elsewhere to port from legacy values, etc.)

samuelallan72:
@pkulkark I added timestamps to messages for the shortanswer and coaching blocks, because one of the original requirements for this was ordering the chats by timestamp. Please check how this fits with other requirements and general consistency. :)

ai_eval/base.py
```diff
@@ -553,3 +582,75 @@ def get_llm_response(self, messages, tag: str | None = None):
...
+        except:
+            logger.warning(
+                "Failed to retrieve location hierarchy information, "
+                "possibly not running in openedx-platform runtime or from within a course. Skipping."
+            )
+            return None
+
+        url_base = get_xblock_app_config().get_site_root_url()
+
+        # XXX: this doesn't work from the CMS, so generating pdfs currently fail from studio (ie. if you interact with the student view in the studio interface).
```
Kelketek: Can we make sure there's a reasonable indicator on the frontend that this isn't available and that they need to view it in the LMS?

ai_eval/coach.py
Kelketek: Would the below code be simpler if the messages were annotated with phony timestamps, then all messages were put into sections using the same code, and there was a flag to hide said timestamps afterward?

Kelketek: Even if not, I think it would be good to break down this function into some smaller ones with good names because there's a lot of indents and it becomes difficult to follow.

ai_eval/coding_ai_eval.py
```diff
@@ -443,3 +451,38 @@ def workbench_scenarios():
...
+        formatter = HtmlFormatter(
+            linenos=False,  # line numbers are difficult here - 'inline' breaks code copy/paste, and 'table' doesn't line up properly. None play well with line wrapping.
+            style='xcode',  # https://pygments.org/styles/
+            noclasses=True, # inline styles so we don't need to mess with external stylesheets
+            wrapcode=True,  # use html5 semantic code elements
+            nobackground=True,  # don't add a background; we want to control the background with our own css
+            prestyles='line-height: 1.5em !important;',  # override default inline styling; it sets it to 125% line-height automatically
+        )
+
+        # TODO: handle output for html/css problems (the output doesn't come from judge0; it's simply the rendered code.
```
Kelketek: Rendered from where? If it's not from Judge0, presumably we'd be rendering it, right? Can we make it safe at that point or do we need to resanitize it here?

ai_eval/pdf_generator.py
```diff
@@ -0,0 +1,186 @@
...
+# A shared jinja environment for consistency and potentially caching.
+_JINJA_ENV = Environment(
+    loader=FileSystemLoader(files("ai_eval").joinpath("static/pdf/templates")),
+    autoescape=True,
+    auto_reload=True,
+)
+_JINJA_ENV.filters["pretty_time"] = pretty_time
+
+
+# XXX: The frontend uses dompurify and marked frontend/src/shared/renderMarkdown.ts;
```
Kelketek: How much work is this? If you think the juice is worth the squeeze I'd recommend making it a follow-up PR rather than tackling it here.

ai_eval/pdf_generator.py
```diff
@@ -0,0 +1,186 @@
...
+        data=data.model_dump(), time=datetime.now(UTC), **metadata.model_dump()
+    )
+
+    css = files("ai_eval").joinpath("static/pdf/base.css").read_text(encoding="utf8")
+
+    # See https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#security for security considerations here.
+    # Any user-provided html must be sanitized with _markdown_to_safe_html()
+    # which should avoid attacks involved malicious html, malicious inline css,
+    # unexpected network requests for external resources, or leaking local files.
+    # XXX: There is a known issue where fontTools (used by Weasyprint) log calls spams the logs with invalid formatting errors.
```
Kelketek: What would it take to fix this? Can we set up a special logging rule for Weasylprint that ignores these problems, and include it with the block? Or would that only be possible by changing the platform's global settings?

ai_eval/pdf_generator.py
pkulkark: The markdown content mostly comes from LLM output which we can't really control, and the PDF needs to follow the designs so I think we should only allow basic formatting. I don't think we need to support richer content; we can always add it later if needed.
Instead of subtracting from nh3.ALLOWED_TAGS, how about using an explicit allowlist of just the basic markdown tags? That way we control exactly what gets through, and any unexpected LLM output gets stripped automatically.

ai_eval/coach.py
pkulkark: Unfortunately we do have to worry about backwards compatibility. All 3 XBlocks are deployed in the client's production instance (but we don't have access to it or maintain it for them).

That said, we don't need to keep compat code forever. Here's how I see it:
- For structural data changes (normalizers, legacy field formats, etc.): PR #24 already introduced Pydantic models with lenient validators that normalize old session data into the new structure on read. So we're already in the "compat release" where old data gets migrated through Pydantic when it's accessed. In a follow-up release, we can tighten the validators and drop the lenient fallbacks, since any session accessed after PR #24's deployment will have been re-saved in the new format.
- For timestamps: This is a different situation. `time` is a brand new field that never existed, so there's nothing to migrate. We should keep the no-timestamps fallback. Old sessions will show coach chat first, then workspace chat (no interleaving). In a later release, once enough time has passed that pre-timestamp sessions are no longer relevant, we can drop the fallback and make time required on CoachedMessage.

pkulkark:
> @pkulkark I added timestamps to messages for the shortanswer and coaching blocks, because one of the original requirements for this was ordering the chats by timestamp. Please check how this fits with other requirements and general consistency. :)

The approach looks good. The only thing from consistency point to consider is adding timestamps to the coding block. It's a single submission rather than a multi-turn conversation, so there's no interleaving to sort, but it'd be nice to have timestamps uniformly across all three blocks.

ai_eval/base.py
```diff
@@ -553,3 +582,75 @@ def get_llm_response(self, messages, tag: str | None = None):
...
+        Return None if an error is encountered that indicates the current runtime doesn't support this
+        or the xblock isn't in a standard course hierarchy (eg. a content library).
+        """
+        try:
+            from openedx.core.djangoapps.xblock.apps import get_xblock_app_config
+            unit = self.get_parent()
+            subsection = unit.get_parent()
+            section = subsection.get_parent()
+            course = section.get_parent()
+        except:
```
pkulkark:
```suggestion
        except Exception:
```

ai_eval/base.py
```diff
@@ -553,3 +582,75 @@ def get_llm_response(self, messages, tag: str | None = None):
             tm[tag] = new_thread_id
             self.thread_map = tm
         return text
+
+    def get_pdf_location_nav(self) -> Location | None:
```
pkulkark: This method should go in compat.py file. That module is specifically the compatibility layer for Open edX platform internals.

ai_eval/coach.py
```diff
@@ -1237,3 +1245,190 @@ def _evaluator_messages():
...
             "finished": session["finished"],
         }
+
+    @XBlock.handler
+    def download_pdf(self, data, suffix=""):
+        """
+        Generate and download the pdf summary of the exercise.
+        """
+        if not self.pdf_download_allowed:
+            raise JsonHandlerError(400, "PDF download is disabled.")
```
pkulkark: Doesn't `JsonHandlerError` require `@XBlock.json_handler`?

pkulkark (changes requested):

ai_eval/pdf_generator.py
samuelallan72: That sounds reasonable @pkulkark , thanks. What did you have in mind for the smaller set of allowed tags for markdown? I originally started writing a short explicit list, but then realised that the default ALLOWED_TAGS  from nh3 is already pretty strict:

```
>>> nh3.ALLOWED_TAGS
{'ins', 's', 'sup', 'code', 'table', 'time', 'h4', 'dt', 'map', 'hr', 'abbr', 'h2', 'tbody', 'rtc', 'q', 'b', 'figcaption', 'colgroup', 'li', 'ol', 'a', 'h6', 'tr', 'ul', 'h5', 'data', 'center', 'figure', 'col', 'thead', 'sub', 'strong', 'kbd', 'del', 'p', 'h1', 'h3', 'nav', 'wbr', 'mark', 'hgroup', 'footer', 'th', 'area', 'span', 'pre', 'img', 'rp', 'acronym', 'div', 'i', 'tt', 'caption', 'rt', 'cite', 'ruby', 'small', 'em', 'header', 'bdo', 'samp', 'strike', 'summary', 'dl', 'var', 'td', 'br', 'u', 'blockquote', 'dd', 'dfn', 'bdi', 'article', 'details', 'aside'}
```

So maybe we could leave it as-is? None of the allowed tags should cause any formatting issues, apart from maybe the tables (but tables may be desired for some llm outputs - I'm thinking for displaying grades, etc.?).

ai_eval/coach.py
samuelallan72: @Kelketek @pkulkark 

> Would the below code be simpler if the messages were annotated with phony timestamps, then all messages were put into sections using the same code, and there was a flag to hide said timestamps afterward?

Oh I like this! This also fits with the idea of normalising old data on read so that existing data is gradually auto-updated to the new format.

> Unfortunately we do have to worry about backwards compatibility. All 3 XBlocks are deployed in the client's production instance

Thanks for the confirmation, good to know. :+1:

ai_eval/base.py
samuelallan72: Yeah I copped out here because I wasn't sure what kind of exceptions would be raised...

ai_eval/coach.py
samuelallan72: Not as far as I know, although I'll retest this. I think the json_handler decorator means you can return data which will be formatted as json data in the response body. But here we want a bare handler because we're returning a custom response object.

ai_eval/pdf_generator.py
samuelallan72: Yeah I think it's worth it, especially to ensure that messages in the frontend look the same as what will be produced in the pdf. Agreed that a follow up PR is a good idea. :+1:

ai_eval/pdf_generator.py
samuelallan72: I spent a couple of hours trying to figure this out... it's tricky because it's coming from the fontTools dependency of Weasyprint. I tried changing log levels for it, deleting the handlers, etc. But I still got that error.

I'm not sure what is causing it though, because the log messages were still successfully printing to the console, after the large error traceback. :/

Maybe I can take another look in a follow up?

ai_eval/coding_ai_eval.py
samuelallan72: Yep we'll be rendering it. So it's tricky here because we're rendering through to pdf, so we should be stricter with sanitising. Although it's probably not too bad - I think the main thing would be to be careful how url downloading is handled to avoid the build crashing or hanging, or internal resources being requested.

But then the flip side is that we need to be also more lenient with sanitising, otherwise the custom html/css will be messed up. And we'd need to check if iframes or other scoping works fine in the pdf generator, otherwise things could leak into the rest of the doc...

I can take a look; not sure how much time to spend on it right now.

ai_eval/coach.py
samuelallan72: > or structural data changes (normalizers, legacy field formats, etc.): PR #24 already introduced Pydantic models with lenient validators that normalize old session data into the new structure on read.

It would be great if these validators normalised the data more completely - currently it only normalises the top level keys, not the values in those keys (eg. right now I can't use this to normalise to adding the time field). Maybe a good follow-up PR :)

ai_eval/coach.py
samuelallan72: Ah indeed you're right. I'll update these to return Responses directly :+1: Nice catch!

ai_eval/coach.py
samuelallan72: I updated this to use fallback times, and it indeed made the logic much simpler :ok_hand:

ai_eval/times.py
```diff
@@ -0,0 +1,33 @@
+from datetime import datetime, UTC, timezone
+
+# These are fallback times to make it easier to work with and order old data that doesn't have times.
+# coach messages should be displayed before workspace messages
+# For any message with these times, the times should not be displayed to the user (because they're fake times).
+FALLBACK_COACH_MESSAGE_TIME = datetime(1975, 1, 1, tzinfo=UTC).isoformat()
+FALLBACK_WORKSPACE_MESSAGE_TIME = datetime(1976, 1, 1, tzinfo=UTC).isoformat()
+
+
+def get_tz() -> timezone:
```
samuelallan72: I factored this out, because I figured it might be desirable to record and display times in a specific timezone (eg. if the client and the lecturers are in Sydney, they may prefer all times to be displayed in Sydney time). Currently everything is just in UTC for simplicity.

Do you think we need to add a setting now (or hook into some existing openedx setting) for preference for timezone?

 @Kelketek @pkulkark

samuelallan72:
Example generated pdf files at this point:

[coached-ai-evaluation-samuel-transcript.pdf](https://github.com/user-attachments/files/26993398/coached-ai-evaluation-samuel-transcript.pdf)
[my-coding-question-samuel-transcript.pdf](https://github.com/user-attachments/files/26993399/my-coding-question-samuel-transcript.pdf)
[my-short-answer-question-samuel-transcript.pdf](https://github.com/user-attachments/files/26993400/my-short-answer-question-samuel-transcript.pdf)

ai_eval/base.py
samuelallan72: I reworked this to ensure it works from both the cms and lms, so this isn't needed now. :grin:

samuelallan72:
@Kelketek @pkulkark thanks for your reviews! I've responded to your comments. I also added screenshots to the PR description, and uploaded example pdfs to the previous comment. :)

ai_eval/times.py
Kelketek: @samuelallan72 We should use https://docs.djangoproject.com/en/6.0/ref/settings/#std-setting-TIME_ZONE I think.

ai_eval/pdf_generator.py
Kelketek: Maybe they're actual print statements or something? In any case I don't think it should be a dealbreaker here-- a follow up for this should be fine.

ai_eval/coding_ai_eval.py
Kelketek: I'd rather we not release anything we know is likely to have significant security holes, so I'd say go ahead and spend some time on it now. If it looks really big once you dig into it, report back so we can strategize. Maybe we can skip the output portion and just deliver the code and the convo.

samuelallan72: @Kelketek 

> I'd rather we not release anything we know is likely to have significant security holes...  Maybe we can skip the output portion and just deliver the code and the convo.

Yep this is exactly the situation right now. The output is simply not rendered.

I'll take a quick look today though and see how much effort will be needed to securely do this.

ai_eval/times.py
samuelallan72: Ah nice, I did some reading, and django sets the timezone information such that we can simply get a date in the local timezone with `datetime.now().astimezone()` :)

ai_eval/coding_ai_eval.py
samuelallan72: Weasyprint doesn't currently support iframes - https://github.com/Kozea/WeasyPrint/issues/1010 . We might be able to do something using object or embed if that works for html, or some kind of two stage process where we render the html to an image, then embed the image in the pdf. Might be good for a follow up, as this will take some time.

ai_eval/coach.py
```diff
@@ -1237,3 +1245,143 @@ def _evaluator_messages():
...
+            character_info = self._get_character_data(entry["character_index"])
+            kind = character_info["pane"]
+            if not sections or sections[-1].kind != kind:
+                sections.append(CoachedSection(kind=kind, messages=[]))
+                if (
+                    kind == "coach"
+                    and self.coach_initial_message
+                    and not used_coach_initial_message
+                ):
+                    sections[-1].messages.append(
```
samuelallan72, line 1320: Adding the initial canned messages here like this means that if the student doesn't interact with one of them, the initial message won't be included in the output. This is possibly a bug. I can address it later if we want the output regardless of whether the student interacts.

samuelallan72:
@pkulkark can we schedule a task to convert the requirements to pyproject.toml format please, and to update the dependencies? It's painful updating all the requirements/*.txt to add new dependencies without causing unwanted updates here.

ai_eval/coach.py
Kelketek: Since the function here is that the learner is going to download their PDF to turn in, rather than some automated submission, I think we can ignore that.

ai_eval/pdf_generator.py
```diff
@@ -127,52 +126,70 @@ def template_file() -> str:
...
     def template_file() -> str:
         return "coding.html"
 
 
-# This data is the same for all templates; it does not include the actual content (chat messages, etc.)
 class Metadata(BaseModel):
+    """Metadata for the PDF.
+
+    This data structure is the same for all PDF templates,
+    and is mostly for data to be included in the metadat, header, or nav.
```
Kelketek: s/metadat/metadata/

Kelketek (commented):
@samuelallan72 This is looking good to me now. Let me know once it's ready in the test environment so I can give it the final thumbs up.

ai_eval/pdf_generator.py
pkulkark: I was thinking of keeping it really basic, like `{'p', 'br', 'b', 'strong', 'i', 'em', 'ul', 'ol', 'li', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'code', 'pre', 'blockquote', 'hr'}` but if the allowed_tags are already strict, I don't see a point in restricting it further. Okay to keep it as-is.

Kelketek:
:+1:

- [x] I tested this: The PDF generation works for me on both Ulmo and Redwood.
- [x] I read through the code
- [x] I checked for accessibility issues (note: It's still a PDF, and there's only so much you can do with those)
- [x] Includes documentation
- [ ] ~Added to the [Code Drift](https://github.com/orgs/open-craft/projects/1) project board (for backports)~

PR #27, opened 2026-05-01
[BB-10772] Refactor Short answer XBlock to be compatible with Ulmo

pkulkark:
## Description

This PR refactors the AI XBlocks to make it compatible with Ulmo. It involves the following:

- upgrade React and Paragon versions
- refactor short answer so it uses paragon styling tokens

## Supporting information

OpenCraft Internal Jira task: [BB-10772](https://tasks.opencraft.com/browse/BB-10772)

## Testing instructions

- Deploy this branch on an Ulmo devstack
- Create a unit and add a short answer xblock component
- Configure it and verify that the studio editor works correctly
- Test learner view and verify the xblock works correctly


ai_eval/static/css/chatbox.css
```diff
@@ -1,3 +1,35 @@
+.shortanswer-react-app {
+  --shortanswer-user-bg: var(--pgn-color-primary-100, #f0f3f5);
```
Kelketek, line 2: @xitij2000 Wanted to check with you since I remember this coming up before in our conversations with WGU-- do we need to make sure that all variables that we're introducing here start with `pgn` in order for them to be overridable by the design tokens build system?

ai_eval/shortanswer.py
```diff
@@ -61,6 +61,15 @@ class ShortAnswerAIEvalXBlock(AIEvalXBlock):
         scope=Scope.settings,
     )
 
+    skip_question = Boolean(
+        display_name=_("Skip Question"),
```
Kelketek: Maybe `hide question` instead? Skip makes it sound like the learner can elect not to answer when it might otherwise be required.

frontend/webpack.config.js
```diff
@@ -56,6 +56,17 @@ module.exports = function webpackConfig(_, argv) {
                 postcssOptions: {
                   plugins: [
                     [
+                      // Scopes Paragon v23 :root { --pgn-* } design-token
```
Kelketek: @pkulkark The idea is to allow the block to inherit the system token theme. Is this preventing that by default?

Kelketek (changes requested):
Hi @pkulkark -- had a change request and a couple of questions.

frontend/webpack.config.js
pkulkark: Ah sorry! That was actually added for Redwood and was supposed to be removed for Ulmo. I've cleaned it up now.

ai_eval/shortanswer.py
pkulkark: Good call. Changed it to `hide question`.

Kelketek:
Thanks, @pkulkark . It's looking good to me now, but I'm currently running a custom theme with Paragon locally. The paragon variables aren't populating within the iFrame so the colors aren't propagating to the block. I don't think that Xblocks automatically have access to the tokens from the Paragon configuration, but @xitij2000 may have an idea how to inject them if you do not.

I'd also like his answer on the token naming question before we merge here.

ai_eval/static/css/chatbox.css
xitij2000: Yes, but hopefully there is going to be a mechanism to override other types soon. For now the approach would be to define these CSS variables separately in our theme and set the value there, i.e. in the customise file we can add:

```css
:root {
  --shortanswer-user-bg: var(--pgn-color-primary-200);
}
```
We can't set this via  design tokens, however we can add a CSS file like the above to bridge from the tokens we have to these variables. 

Using a custom css variable is the recommended approach, here  by default `shortanswer-user-bg` will pull from the paragon theme, but here is one issue which is that this is too specific. i.e. if we want to customise this variable, we can't seet it using the above approach in the theme, since that is in the root scope, and the variable defined in the `shortanswer-react-app` class is more specific, so to override it from the theme it wold also need to be set in the `.shortanswer-react-app` class. If we want to be able to set it without needing to figure out where it's defined we should define it at `:root`.

I don't know the intention of this CSS variable. If we're using the same value multiple times it might make sense to introduce a "local" variable like this, without the intention of ever overriding it.

xitij2000:
> I don't think that Xblocks automatically have access to the tokens from the Paragon configuration, but @xitij2000 may have an idea how to inject them if you do not.

Yes, they don't. I'm working on adding this support for the blocks in https://gitlab.com/opencraft/dev/exemplar-blocks so that can be used as reference for how to do it.

xitij2000:
@Kelketek @pkulkark  Here is the PR with the runtime theme loading: https://gitlab.com/opencraft/dev/exemplar-blocks/-/merge_requests/24

Currently it doesn't apply the theme properly on studio since loading the theme properly requires knowing the MFE_CONFIG_API URL which XBlocks don't. So what I'd need to do is pass the LMS url to the XBlock so it knows where to get the MFE_CONFIG_API and then load that. 

Alternatively we can just load the themes via the backend an pass the URLs directly to the block.

xitij2000:
@Kelketek @pkulkark Actually wait a bit, there are complications here with the shadow dom that are causing the styles to not apply properly. I will disable it for now and figure that out later.

ai_eval/static/css/chatbox.css
pkulkark: Moved it to `:root`.

Kelketek (approved):
:+1:

- [x] I tested this: New UI works as expected, looks as expected.
- [x] I read through the code
- [x] Includes documentation: new field is documented, other changes are visual
- [ ] -Added to the [Code Drift](https://github.com/orgs/open-craft/projects/1) project board (for backports)-

Kelketek:
Per our conversation on Jira, actually loading the tokens can be done in a follow-up.

ai_eval/static/css/chatbox.css
xitij2000: There is already a PR for this: https://github.com/openedx/paragon/pull/4276

PR #28, opened 2026-05-01
docs: add claude API key docs

Kelketek:
## Description

Make it clear how to set the CLAUDE_SONNET API key.

## Testing instructions

1. Install the block.
2. Set the API key in the site configuration according to the directions.
3. Create a block and select Claude Sonnet
4. Verify that the block picks up the key and locks the key field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ The system will check for these configurations in the following order:
     {
         "ai_eval": {
              "GPT4O_API_KEY": "your-openai-api-key",
+             "CLAUDE_SONNET_API_KEY": "your-anthropic-api-key",
              "LLAMA_API_URL": "https://your-llama-endpoint"
         }
     }
@@ -104,6 +105,7 @@ The system will check for these configurations in the following order:
    XBLOCK_SETTINGS = {
        "ai_eval": {
            "GPT4O_API_KEY": "your-openai-api-key",
+           "CLAUDE_SONNET_API_KEY": "your-anthropic-api-key",
            "LLAMA_API_URL": "https://your-llama-endpoint"
         }
     }


pkulkark (approved):
LGTM 👍

Kelketek:
@pkulkark It looks like CI is broken on the main branch. Is there already some work in flight for this? It seems the issue is that the image for 20.04 isn't available anymore. I tried bumping the image like [in this PR in the ulmo branch](https://github.com/open-craft/xblock-ai-evaluation/commit/03a8a78d07fdf07e93a2a0165804e0674978a676#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR21), but this broke things in other ways.

I'm considering merging anyway since this is docs-only, but we should fix it, so let me know if I should go ahead and try fixing this now or create a follow-up ticket for scheduling.

samuelallan72:
> It looks like CI is broken on the main branch. Is there already some work in flight for this?

@Kelketek cc @pkulkark just a drive-by comment: yes, the main branch is essentially abandoned. The current branch that we've been merging to, and is used in production, is [`artur/all-changes`](https://github.com/open-craft/xblock-ai-evaluation/tree/artur/all-changes).

Is anything stopping us from merging artur/all-changes to main, and switching the production branches to use main? That would be neater and less confusing.

Kelketek:
I'd certainly prefer that, @samuelallan72 . Any reason against, @pkulkark ?

samuelallan72:
@Kelketek I don't know the ticket for this, but just checking in since it's been a while. If there are no preferences, I'd recommend rebasing this on the artur/all-changes branch and merge it. Then we can figure out the question of main vs artur/all-changes later.

@pkulkark could you check this please?

Kelketek:
@samuelallan72 I'm pretty sure that WGU will be on Ulmo soon (they're working on it now) so I think it would actually be better to put this on `opencraft/ulmo` and then once WGU is on Ulmo, replace master with that branch and use it moving forward. I'm pretty sure feature-wise, artur's branch and the Ulmo branch are in sync.

Any objections to this, @pkulkark ?

PR #29, opened 2026-05-04
fix: Pin weasyprint version to Redwood compatible one

pkulkark:
## Description

This PR pins the weasyprint version to `>=57.2,<62` to fix the tinycss2 version mismatch when the XBlock is installed which results in Studio service erroring out.

The main openedx platform constraints tinycss2 to version 1.2.1 for Redwood and 1.4.0 for Ulmo. Weasyprint versions 57.2-61.x use tinycss2 >= 1.0.0, 62.x bumps it to >=1.3.0 and 63.x requires > 1.4.0, making the range `>=57.2,<62` compatible with Redwood and Ulmo.

## Supporting Information

OpenCraft Internal Jira task: [BB-10694](https://tasks.opencraft.com/browse/BB-10694)


requirements/base.txt
```diff
@@ -176,69 +187,73 @@ pydantic==2.11.7
...
-    #   anyio
-    #   openai
-sqlparse==0.5.3
+    # via openai
+sqlparse==0.5.5
     # via django
-tiktoken==0.11.0
+tiktoken==0.12.0
     # via litellm
 tinycss2==1.5.1
```
samuelallan72, line 245: @pkulkark could you confirm this comment wasn't the case in the end?

> it seems it takes the latest available dependency for tinycss2: (tinycss2>=1.5.0 https://pypi.org/project/tinycss2/1.5.1/) and the redwood version we are running has dependencies that expects version tinycss2 1.2.1

Because there's tinycss2 1.5.1 here still. :)

setup.py
```diff
@@ -35,7 +35,7 @@ def package_data(pkg, roots):
         "celery",
         "chardet",
         "litellm",
-        "weasyprint",
+        "weasyprint>=57.2,<62",
```
samuelallan72, line 38: @pkulkark do you know which set of requirements is used when installing the xblock into an instance? I would have thought it would use this install_requires list here, and ignore the `requirements/*.txt` right? Because it would be installed with pip? The `requirements/*.txt` are only used for tests?

samuelallan72:
@pkulkark please update the PR description to explain the changes (including why pinning to this particular weasyprint version fixes compatibility) and link to the Jira ticket. :)

requirements/base.txt
pkulkark: Confirmed it on my redwood devstack. The requirements/*.txt files are only used for tests/CI. During installation, it only uses `install_requires`.

setup.py
pkulkark: @samuelallan72 Correct. While installing, pip only uses install_requires from setup.py. I added the same in base.in and ran `make upgrade` which resulted in the other requirements/*.txt updates.

ai_eval/pdf_generator.py
```diff
@@ -190,6 +190,5 @@ def generate_pdf(
         # we can put css in a separate file, or include it in a <style> element in the template html
         stylesheets=[CSS(string=css)],
         # archival and accessible standard - see https://pdf.abbyy.com/learning-center/pdf-standards/
```
samuelallan72: @pkulkark just so we don't forget in future:

```suggestion
        # archival standard - see https://pdf.abbyy.com/learning-center/pdf-standards/
        # NOTE: when we update to weasyprint >= 67.0, change this variant to pdf/a-3a, with `pdf_tags=True` for improved accessibility.
        #       https://doc.courtbouillon.org/weasyprint/stable/changelog.html#version-67-0
```

requirements/base.txt
samuelallan72: Ah I guess it didn't pin tinycss2 back to the older version is that this doesn't include enough constraints from the openedx platform... hmmm.

We really should delete all these requirements/* files and add a pyproject.toml. These files are a pain to maintain, and if these files aren't even used when installing the xblock and they don't pick up enough constraints...

samuelallan72 (approved):
@pkulkark thanks for updating the PR description, that makes sense. I left a suggestion about updating the comment on the pdf standard, but otherwise fine to me. :)

PR #32, opened 2026-05-20
feat: support runtime theme styles

samuelallan72:
## Description

Port from https://gitlab.com/opencraft/dev/exemplar-blocks/-/merge_requests/34

Dynamically load the runtime theming for use by the block, so the blocks will be also styled by the theme.
This also means we can use the paragon and theme css variables in the custom css here, supporting further dynamic theming in future.

## Supporting information

Private-ref: https://tasks.opencraft.com/browse/BB-10816

## Testing instructions

- deploy this and create an instance of each of the three ai evaluation xblocks
- verify the blocks still work fine
- verify the body text now uses the expected font (Lato) - this was a high priority fix I included with this - further style updates can happen in follow up PRs once this is merged.


frontend/src/shared/mountApp.tsx
```diff
@@ -16,55 +59,87 @@ function toDomElement(element: XBlockElementLike) {
...
-function resolveMountNode(elementLike: XBlockElementLike, selector?: string) {
-  const element = toDomElement(elementLike);
-  let mountNode: Element | null = null;
-
-  if (selector) {
-    mountNode = element.querySelector(selector);
-    if (mountNode instanceof HTMLElement) {
-      return mountNode;
-    }
+const ENABLE_SHADOW_ROOT = true;
```
samuelallan72, line 62: I kept this toggle, although I have not tested the false path, since we still use that toggle in exemplar blocks. Not sure if we can drop it yet?

frontend/src/shared/styles.css
```diff
@@ -7,13 +7,19 @@
...
   color: inherit;
   background-color: transparent;
 }
 
+.ai-eval-paragon {
+    /* override injected styles from lms-course */
+    p, label {
+        font-family: var(--pgn-typography-font-family-base) !important;
+    }
+}
```
samuelallan72, line 22: These are the styles added to fix the body text font family. I can split into a separate PR if preferred.

frontend/src/shared/mountApp.tsx
xitij2000: I kept it as a quick escape hatch in case we need to switch back. Let's keep it around for now and we can do a clean up and remove it later if it is no longer needed.

ai_eval/base.py
```diff
@@ -504,6 +505,8 @@ def _build_view_payload(
             "handler_urls": handler_urls,
             "initial_state": initial_state,
             "meta": meta,
+            "mfe_config_api": f"{settings.LMS_ROOT_URL}/api/mfe_config/v1?mfe=learning",
+            "style_urls": [settings.LMS_ROOT_URL + self.runtime.local_resource_url(self, url) for url in style_urls],
```
xitij2000, line 509: Rather than passing these, we can import them since the URLs are fixed.

xitij2000: Actually, I'm not sure why these are not in the frontend directory and directly loaded there. Could they be moved?
cc: @pkulkark

xitij2000 (commented):
The code generally looks good! I wonder if we can consolidate the CSS, since that would passing it from the backend to the frontend. It's a static list of assets so if nothing else the list can be moved to the frontend instead of the backend. 

I'll just test it soon and approve the PR if it's working as expected.

Kelketek:
@xitij2000 I've tested it and it works. You mentioned you'd approve it based on testing, and I think the other notes are minor and can be addressed in a follow-up PR where Sam is doing the remaining styling work, so I'm going to go ahead and merge now.

samuelallan72:
Thanks @Kelketek @xitij2000 ! :)

PR #33, opened 2026-05-21
Update shortanswer styling to match new designs

samuelallan72:
## Description

Update the shortanswer ai chat to match the [new designs](https://www.figma.com/design/lxSfiMgNARbxmwkCHf8Wdj/Course-Template-Designs?node-id=1383-8469&t=fkmItxZeuDQUEvd9-0) (private link).

<details>

image

image
</details>

## Supporting information

## Testing instructions

- deploy this branch, add a new shortanswer ai component, configure, and send some messages
- compare with the designs and verify they match
- check the other ai blocks and verify the styles are not messed up

## known issues

- I discovered a double scrolling problem with the studio editor for the short answer block - we probably should address that sometime
- We don't have new designs that incorporate the image field, so I'm not sure how that should look. I've tentatively put it under the new question text so it looks a bit nicer, but it's a bit janky. Let me know if you reckon it should go elsewhere, otherwise we can address it in a followup.


ai_eval/static/css/chatbox.css
xitij2000, line 1: This file has a few hardcoded colors. Other than that all seems good.

samuelallan72: Thanks @xitij2000 :) I think this will go through a lot of changes though, since I need to rebase on opencraft/ulmo.

xitij2000: Sure! I'll have a look at it again once it's ready.

ai_eval/static/bundles/shared.css
```diff
@@ -10,6 +10,10 @@
   --ai-eval-body-font: var(--pgn-typography-font-family-base);
 }
 
+.text-secondary-500 {
+    color: var(--pgn-color-secondary-500);
```
samuelallan72: @xitij2000 I noticed that there are no utility colour classes based on the secondary colour. Maybe this is something we can add an upstream?

xitij2000: I think these already exist upstream. perhaps they are not documented, but they definitely exist.

Kelketek: @samuelallan72 Were you able to find these utility color classes as @xitij2000 mentioned?

ai_eval/static/css/chatbox.css
```diff
@@ -30,6 +30,8 @@
   --shortanswer-button-font-size: var(--pgn-typography-font-size-sm, 0.875em);
   --shortanswer-button-line-height: var(--pgn-typography-line-height-sm, 1.25em);
   --shortanswer-submit-button-gap: var(--pgn-spacing-spacer-1-5, 0.35rem);
+
+  --shortanswer-message-user-bg-color: #E9F6F5;
```
Kelketek: @samuelallan72 This one is still hardcoded.

ai_eval/static/css/chatbox.css
```diff
@@ -38,6 +40,29 @@
   clear: both;
 }
 
+.shortanswer-question {
+    background-color: var(--pgn-color-gray-200);
+    border-radius: 16px;
```
Kelketek: This is still hardcoded.

Kelketek (changes requested):
@samuelallan72 I'm getting some weird effects here. What I see in studio:

Image

And here's what it looks like for me in the LMS:

Image

Note the weirdness with the input.

One thing that I'm unsure of is whether the question should be collapsed by default. I'm thinking 'no', since there could be significant info in that area initially.

ai_eval/static/bundles/shared.css
samuelallan72: @xitij2000 ah nice, yes they indeed exist, just not documented :ok_hand:

ai_eval/static/css/chatbox.css
samuelallan72: Addressed!

ai_eval/static/css/chatbox.css
samuelallan72: Addressed! :)

ai_eval/static/bundles/shared.css
samuelallan72: Addressed! :)

samuelallan72:
@Kelketek 

> One thing that I'm unsure of is whether the question should be collapsed by default. I'm thinking 'no', since there could be significant info in that area initially.

Agreed, sounds good. I've made it default to expanded. :+1: 

> Note the weirdness with the input.

I couldn't reproduce this - it may be something conflicting in your devstack (as discussed on mattermost). :thinking: 

For the studio styling - I see this too, but it's not related to this PR. I'll take more of a look and maybe we can do a follow up PR? (I can try to get to this tomorrow.)

Kelketek:
@samuelallan72 

> Agreed, sounds good. I've made it default to expanded. 👍

I started thinking about this a little more, and I think there's actually a subtly best way of doing this that's slightly more complex but likely better, product-wise.

I think the panel should be expanded on first view. Once the user submits a response, it should collapse and stay collapsed by default from that point forward. So, if loading the block later, and there's a response from the LLM, it should be collapsed. If not, it should be open.

The issue with the form styling seems to be on the live instance:

image

Could you take a look at that?

PR #36, opened 2026-05-25
refactor: add shadow dom support (ulmo branch)

samuelallan72:
## Description

Rebase the changes from https://github.com/open-craft/xblock-ai-evaluation/pull/32

This should help with isolation between the xblocks styles and other styles from the platform.

## Supporting information

## Testing instructions

- verify everything builds
- add the blocks to a course and verify they render as expected. Rendering should be the same before and after this PR. NOTE: I've only tested the shortanswer xblock here.


ai_eval/base.py
```diff
@@ -461,12 +461,13 @@ def _apply_studio_issues(
 
     # Frontend payload helpers
 
-    def _build_view_payload(
+    def _build_view_payload(  # pylint: disable=too-many-positional-arguments
```
Kelketek: Any reason we can't make these keyword arguments, rather than positional? I think the linter is right here, barring some technical reason.

In fact, below it looks like we're invoking as keyword arguments anyway. You should be able to force these as required kwargs like so:

```python
    def _build_view_payload(
        self,
        *,
        view: str,
        handler_urls: dict[str, str],
        initial_state: dict[str, Any],
        meta: dict[str, Any],
        style_urls: list[str],
    ) -> dict[str, Any]:
```

Kelketek (commented):
Started reviewing, but then realized this PR is in draft state, so stopped-- let me know once it's ready.

ai_eval/base.py
samuelallan72: @Kelketek no reason, I just continued with how it was originally used. I can change it :)

samuelallan72: Updated!

frontend/src/shared/mountApp.tsx
xitij2000, line 1: Could we copy the utils directly from the exemplar blocks? If it's the same file with minor modifications, it will be easier to diff and copy over changes in the future.

xitij2000 (approved):
The code looks good! Just one minor suggestion to make it easier to port changes later. 

I'm approving on the basis of the code, but will test it and leave another approval.

frontend/src/shared/mountApp.tsx
samuelallan72: @xitij2000 that sounds like a good idea. Are you ok if this is done in a future refactoring PR though, to unblock working on the new styles sooner? :)

xitij2000: That's definitely okay. This isn't a blocker, unless it doesn't work in testing which I'm doing right now :-)

xitij2000 (approved):
This is good to merge! Tested on devstack.

PR #40, opened 2026-05-28
feat: improve question panel collapsible behaviour

samuelallan72:
## Description

For the question panel in the AI shortanswer block, this makes the behaviour a bit more dynamic and convenient:

- For the empty state (no messages), the panel is expanded by default. This is so the student can see all important instructions at first, so they don't miss anything.
- When a user sends their first message after an empty state, the panel is auto-collapsed. This is so less screen realestate is used, and it's likely the student already knows the instructions at this point.
- When sending further messages, the panel state is not changed. This is to avoid changing a user's preference later during a conversation, which could be annoying.
- When loading the page when there are already messages, the panel is collapsed by default. This is to make better use of the screen realestate, and it's likely the student already knows the instructions at this point.
- When the chat is reset, the panel is auto-expanded. This is to remind the student of the instructions before re-trying the problem.

## Supporting information

## Testing instructions

- deploy this branch and add an instance of the ai shortanswer block to a course
- interact with the block and verify the behaviour is as described above
- verify the behaviour makes sense


Kelketek:
@samuelallan72 Gave this a try. It works, but I have a few notes:

1. The number of cases when we're manipulating the state certainly necessitates tests. Please add them.
2. I noticed after I write a multi-line response to the LLM and submit it, the textarea remains expanded. It should probably shrink back after emptying. Is there something we can do here?
3. The description mentions that the question should re-expand 'after the chat is reset'. How do I reset the chat? I don't see an option for that.

samuelallan72:
@Kelketek 

> The description mentions that the question should re-expand 'after the chat is reset'. How do I reset the chat? I don't see an option for that.

Ah sorry, you need to enable the "Allow reset" option in the studio editor for the shortanswer ai block:

image

> I noticed after I write a multi-line response to the LLM and submit it, the textarea remains expanded. It should probably shrink back after emptying. Is there something we can do here?

Ah good catch, I can check that.

>  The number of cases when we're manipulating the state certainly necessitates tests. Please add them.

Sure I can have a go at adding something.

samuelallan72:
On the topic of the reset button, I didn't think of that for multi line text, so that probably needs tidying:

image

samuelallan72:
>    I noticed after I write a multi-line response to the LLM and submit it, the textarea remains expanded. It should probably shrink back after emptying. Is there something we can do here?

I think this is a bug in Paragon or react-bootstrap - the auto-resize handler doesn't trigger when the value changes; it's only triggered from the "onChange" handler. Setting the controlled value mustn't trigger the "onChange". I'll think about how we might address that.

samuelallan72:
I've fixed the textarea autoresize issue, and also fixed the reset button growing too tall:

image

samuelallan72:
Ah I haven't added any tests yet, but I've run out of time today. I can add them in a follow up later if that's ok? @Kelketek

Kelketek:
@samuelallan72 Yes, that works for me-- please do so in the morning.

:+1:

PR #42, opened 2026-06-01
feat: vertical spacing variable

Kelketek:
## Description

This PR adds a vertical spacing variable to the shortanswer AI block, allowing us to add in spacing where it's missing in a theme that collapses the gap between blocks.

## Supporting information

https://tasks.opencraft.com/browse/BB-10840

## Testing instructions

1. Install this block
2. Add a short answer block
3. Set up [the simple theme dev server from this branch](https://github.com/open-craft/edx-simple-theme/pull/75), according to the [README](https://github.com/open-craft/edx-simple-theme/tree/wgu-release/ulmo#developing-a-new-theme)

### DIFF
--- a/ai_eval/static/bundles/shared.css
+++ b/ai_eval/static/bundles/shared.css
@@ -1,13 +1,46 @@
 /* use root and host selectors to work both in and out of the shadow dom,
  * for if we toggle use of the shadow dom later */
 :root, :host {
-  --ai-eval-react-spacing-sm: 0.5rem;
-  --ai-eval-react-spacing-md: 1rem;
-  --ai-eval-react-border-color: #d0d0d0;
-  --ai-eval-react-surface-color: #ffffff;
-  --ai-eval-react-muted-color: #5c5c5c;
-  --ai-eval-react-radius: 0.5rem;
-  --ai-eval-body-font: var(--pgn-typography-font-family-base);
+  @property --ai-eval-react-spacing-sm {
+    syntax: "<length>";
+    inherits: true;
+    initial-value: 0.5rem;
+  }
+  @property --ai-eval-react-spacing-md {
+    syntax: "<length>";
+    inherits: true;
+    initial-value: 1rem;
+  }
+  @property --ai-eval-react-border-color {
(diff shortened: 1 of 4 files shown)


ai_eval/static/bundles/shared.css
```diff
@@ -1,13 +1,46 @@
...
+  }
+  @property --ai-eval-body-font {
+    syntax: "*";
+    inherits: true;
+    initial-value: var(--pgn-typography-font-family-base);
+  }
+  @property --ai-eval-vertical-padding {
+    syntax: "<length>";
+    inherits: true;
+    initial-value: 0;
```
Kelketek, line 42: @xitij2000 Calling your attention to this. This is the first time we've created a unique variable we intended to immediately override. All the other ones pick up from a paragon default or have a default constant value we're cool with.

In this case we want the default spacing to be zero because this override shouldn't be needed in the default theme. However, since our theme collapses the space between blocks this is needed.

What I discovered was that the way we were setting default values makes them ignore the properties set in the theme. They must be defined via `@propery` and have an `initial-value` rather than defining the property's exact value as we were doing before.

ai_eval/static/bundles/shared.css
```diff
@@ -1,13 +1,46 @@
 /* use root and host selectors to work both in and out of the shadow dom,
```
pkulkark, line 1: @Kelketek Shouldn't these changes be made in the `frontend/src/shared/styles.css` file and then build the bundles rather than making the changes in the bundled css directly?

Kelketek: @pkulkark Yep, looks like I edited the wrong file. I'll fix it.

Kelketek:
@pkulkark Addressed your notes. Remembered to run a rebuild as well, and discovered that a package was missing for Typescript (uuid's types package) so that has been added as well.

pkulkark (approved):
LGTM 👍 

- [x] I tested this: Verified the spacing variable is being overridden through the theme as described in the testing instructions
- [x] I read through the code

PR #43, opened 2026-06-04
Update frontend tests

samuelallan72:
## Description

- add CI for frontend tests
- fix the frontend tests

## Supporting information

Private-ref: https://tasks.opencraft.com/browse/BB-10839

Follow up of #40 

## Testing instructions

- verify CI passes
- verify frontend tests are correctly run in CI
- verify new tests successfully verify the behaviour added in #40 
- verify only tests and CI are changed; no app code should be changed

### DIFF
--- a/frontend/jest.config.js
+++ b/frontend/jest.config.js
@@ -14,6 +14,6 @@ module.exports = {
     }],
   },
   transformIgnorePatterns: [
-    "/node_modules/(?!(@openedx/paragon|@openedx/paragon/icons|dompurify|marked)/)",
+    "/node_modules/(?!(@openedx/paragon|@openedx/paragon/icons|dompurify|marked|uuid)/)",
   ],
 };
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Python CI
+name: Tests
 
 on:
   push:
@@ -13,12 +13,11 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  tests:
-    runs-on: ${{ matrix.os }}
+  backend-tests:
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04]
         python-version: [3.11, 3.12]
         toxenv: [django42, quality]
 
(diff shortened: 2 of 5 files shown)


frontend/jest.config.js
```diff
@@ -14,6 +14,6 @@ module.exports = {
     }],
   },
   transformIgnorePatterns: [
-    "/node_modules/(?!(@openedx/paragon|@openedx/paragon/icons|dompurify|marked)/)",
+    "/node_modules/(?!(@openedx/paragon|@openedx/paragon/icons|dompurify|marked|uuid)/)",
```
samuelallan72, line 17: This was to fix:

```
    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /home/samuel/proj/opencraft/devstack-master/xblock-ai-evaluation/frontend/node_modules/uuid/dist/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){export { default as MAX } from './max.js';
                                                                                      ^^^^^^

    SyntaxError: Unexpected token 'export'
```

Kelketek:
@samuelallan72 

:+1:

- [ ] ~I tested this: (describe what you tested)~ CI tested, looks good.
- [x] I read through the code
- [ ] ~I checked for accessibility issues~ No new user-facing code introduced
- [x] Includes documentation: Appropriate comments added, no other docs needed here.
- [ ] ~Added to the [Code Drift](https://github.com/orgs/open-craft/projects/1) project board (for backports)~ N/A

Kelketek (approved):

PR #44, opened 2026-06-09
fix: fix typing indicator for shortanswer ai block

samuelallan72:
## Description

The typing indicator for the shortanswer block was not visible due to css making it the same colour as the background. This was probably caused by an oversight during the recent theming work.

This fixes that issue. It also refactors the typing indicator into a separate component, deduplicating code between the shortanswer block and the coach block. I used the animation from the coach block, because that seemed more eye-pleasing and consistent with other common typing indicators I've seen.

## Supporting information

[Slack thread](https://openedx.slack.com/archives/C0AK07RLBEJ/p1780676663644099)

Private-ref: https://tasks.opencraft.com/browse/BB-10901

## Testing instructions

- create an instance of the shortanswer and the coach blocks
- configure and publish
- in each block, send a message in each chat window
- verify that a consistent typing indicator is displayed in the appropriate chat window while you're waiting for the llm bot response


ai_eval/templates/coach_layout.html
samuelallan72, line 1: This file wasn't used at all -  I wonder if we can remove others from the templates dir too.

I removed it because it was confusing seeing multiple places where the typing indicator / spinner was defined.

frontend/src/shared/TypingIndicator.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+// A small animated typing indicator
+export const TypingIndicator = () => {
+  return (
+      <div className="typing-indicator" aria-hidden="true">
```
Kelketek, line 6: This is hidden, but I think we have some kind of indicator for screenreaders, like the text 'thinking...'

Kelketek:
@samuelallan72 I had one note on accessibility-- but if you address that, this is :+1: from me.

frontend/src/shared/TypingIndicator.tsx
samuelallan72: @Kelketek I _think_ this is already addressed by this line: https://github.com/open-craft/xblock-ai-evaluation/blob/ddfc4e61c63b59791f30a276e2187e7333fea3d2/frontend/src/shortanswer/ShortAnswerStudentApp.tsx#L104

Which marks the chat window as busy/loading/thinking for screenreaders.

samuelallan72:
(force pushed to fix commit message)

PR #45, opened 2026-06-16
fix: focus and scroll related issues after short answer submit response

rpenido:
## Description

Fixes two focus-related issues in the short answer student component:
- The send button steals focus on click
- The textarea refocusing after receiving a backend response.

This was causing a UI issue in which the user's scroll was disrupted after the API response.
As a side effect, the edit needs to be enabled after submission (to keep the focus), but it will not submit until the backend responds.

## Testing instructions

First, try to reproduce the bug using the `opencraft/ulmo` branch:
- Add a "Short answer with AI Evaluation" XBlock
- Add a Text block after it with a lot of text so you can scroll past the AI Block
- Go to the "Preview" of the Unit
- Type some text in the AI Xblock input, send it and scroll down to move the AI XBlock out of the view
- After the server return the AI response, the input is selected, moving the current view to the AI XBlock

Then, checkout this branch, repeat the process and check that the focus/scroll doesn't change after the backend result

- Test that the user cannot submit another answer while the server is processing, typing ctrl + enter
___
Private ref: [FAL-4366](https://tasks.opencraft.com/browse/FAL-4366)


navinkarkera:
@rpenido The bug is fixed but now the focus is lost after each user submission.

rpenido:
> @rpenido The bug is fixed but now the focus is lost after each user submission.

@navinkarkera This is a tricky issue, and I'm not sure how best to handle it.
We could call `.focus({ preventScroll: true })`, which would focus again without scrolling, but it would still be problematic for:
- Potentially changing the focus to a component outside the view window
- If the user is typing in another component, changing the focus after the API response would still cause a UX issue

We could also prevent the focus from being lost in the first place, but in that case, we would need to keep the control enabled after submission (which would be another issue).

Sorry for not mentioning this in the PR. I updated it now.

Do you think we could have a better approach here?

navinkarkera:
@rpenido I think it would be better to prevent the focus from being lost in the first place. Keep the control enable but don't allow user to submit.

rpenido:
> I think it would be better to prevent the focus from being lost in the first place. Keep the control enable but don't allow user to submit.

Done @navinkarkera! Fixed here: https://github.com/open-craft/xblock-ai-evaluation/pull/45/changes/8691d22ed09bda2c506f9716b7f520c008d584dd

frontend/src/shortanswer/ShortAnswerStudentApp.tsx
navinkarkera:
```suggestion
      setDraft((current) => current || userInput);
```

If the user has already typed something into the field while waiting for response, we don't want to override it with previous answer.

navinkarkera (approved):
@rpenido Looks good.

PR #46, opened 2026-06-16
[BB-10960] Replace retired Claude Sonnet 4 model with Sonnet 4.6

pkulkark:
## Description
The AI-evaluation XBlocks offered Anthropic's claude-sonnet-4-20250514 (Claude Sonnet 4) as a selectable model. Anthropic has retired that model, so any request to it now fails at the provider and learners get an error instead of an AI response.

This PR:
- Switches the supported Claude model to claude-sonnet-4-6 (Claude Sonnet 4.6), Anthropic's official drop-in replacement.
- Adds a backwards-compatibility shim that resolves the retired id to its replacement at the point of use.

## Supporting Information
OpenCraft's Internal Jira task: [BB-10960](https://tasks.opencraft.com/browse/BB-10960)

## Testing Instructions
1. New activity: Add or edit an AI-eval block in Studio, open settings, confirm the model dropdown lists claude-sonnet-4-6 (and no longer lists claude-sonnet-4-20250514). Select it, provide a valid Anthropic API key, and save.
2. In the LMS, submit a learner response and confirm an AI reply is returned.
3. Legacy activity (the shim): Take a block whose stored model is still claude-sonnet-4-20250514, do not re-select the model, and submit a learner response in the LMS. Confirm it succeeds - the request is served by claude-sonnet-4-6 with no edit to the block.

### DIFF
--- a/ai_eval/supported_models.py
+++ b/ai_eval/supported_models.py
@@ -11,10 +11,24 @@ class SupportedModels(Enum):
     GPT4O = "gpt-4o"
     GPT4O_MINI = "gpt-4o-mini"
     GEMINI_PRO = "gemini/gemini-pro"
-    CLAUDE_SONNET = "claude-sonnet-4-20250514"
+    CLAUDE_SONNET = "claude-sonnet-4-6"
     LLAMA = "ollama/llama2"
 
     @staticmethod
     def list():
         """Return the list of supported model values."""
         return [str(m.value) for m in SupportedModels]
+
+
+# Models retired or renamed by their providers, mapped to their drop-in
+# replacement. Activities saved with a legacy identifier keep working because
+# the value is resolved to its replacement at the point of use (the provider
+# call and the per-model API-key lookup), so no course content needs editing.
+LEGACY_MODEL_ALIASES = {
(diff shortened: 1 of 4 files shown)


ai_eval/supported_models.py
```diff
@@ -11,10 +11,24 @@ class SupportedModels(Enum):
...
         """Return the list of supported model values."""
         return [str(m.value) for m in SupportedModels]
+
+
+# Models retired or renamed by their providers, mapped to their drop-in
+# replacement. Activities saved with a legacy identifier keep working because
+# the value is resolved to its replacement at the point of use (the provider
+# call and the per-model API-key lookup), so no course content needs editing.
+LEGACY_MODEL_ALIASES = {
+    "claude-sonnet-4-20250514": "claude-sonnet-4-6",
```
Kelketek, line 28: @pkulkark Could we make this value the enum instead? That way we don't have to update the values in this dict each time there's a new release-- only add an entry.

Kelketek (approved):
@pkulkark This is working for me. I have one minor note. In the interest of time, please address it in a follow-up first thing in the morning. Thanks!

PR #48, opened 2026-06-16
fix: Use supported model enum instead of string

pkulkark:
Follow-up PR for #46

### DIFF
--- a/ai_eval/tests/test_ai_eval.py
+++ b/ai_eval/tests/test_ai_eval.py
@@ -1793,6 +1793,23 @@ def test_resolve_model_maps_legacy_to_replacement():
     assert resolve_model("some/custom-model") == "some/custom-model"
 
 
+def test_resolve_model_follows_alias_chain():
+    """A chained alias (code default -> operator replacement) resolves to the end of the chain."""
+    # e.g. code: claude-sonnet-4-20250514 -> claude-sonnet-4-6;
+    #      operator DEPRECATED_MODELS:     claude-sonnet-4-6 -> claude-sonnet-4-10
+    aliases = {
+        LEGACY_CLAUDE: SupportedModels.CLAUDE_SONNET.value,
+        SupportedModels.CLAUDE_SONNET.value: "claude-sonnet-4-10",
+    }
+    assert resolve_model(LEGACY_CLAUDE, aliases) == "claude-sonnet-4-10"
+    assert resolve_model(SupportedModels.CLAUDE_SONNET.value, aliases) == "claude-sonnet-4-10"
+
+
+def test_resolve_model_terminates_on_cycle():
+    """A misconfigured alias cycle must not loop forever."""
+    assert resolve_model("A", {"A": "B", "B": "A"}) in {"A", "B"}
+
+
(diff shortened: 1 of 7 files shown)


Kelketek:
@pkulkark One more request, now that I'm thinking on it-- can we make the current Claude model overwritable as a setting, and also make it to where we can have "deprecated claude/chatGPT/etc models" settings so that the replacement dictionary can be updated that way instead of having to make a code change each time?

pkulkark:
@Kelketek Thanks, that's a good idea. I've pushed the changes for it. Please take a look.

ai_eval/tests/test_ai_eval.py
```diff
@@ -1813,20 +1820,90 @@ def test_get_llm_response_calls_provider_with_resolved_model(shortanswer_block_d
...
+    block = ShortAnswerAIEvalXBlock(ToyRuntime(), DictFieldData(shortanswer_block_data), None)
+    with patch(
+        "ai_eval.supported_models.get_site_configuration_value",
+        side_effect=_fake_site_config({"GPT4O_MINI_MODEL": SupportedModels.GPT4O.value}),
+    ):
+        # "gpt-4o" is GPT4O's default; defaults win, so it maps to GPT4O, not GPT4O_MINI.
+        assert block._get_model_config_key(SupportedModels.GPT4O.value, "api_key") == "GPT4O_API_KEY"
+
+
+def test_model_maps_are_cached_per_instance(shortanswer_block_data):
```
Kelketek, line 1916: @pkulkark This only lives per render request, right? It wouldn't cause a problem where the block gets stuck with an old config and requires the container to be restarted to load the revised config?

Kelketek:
@pkulkark This doesn't appear to be working for me.

I'm using the following config:

```json
{
    "ai_eval": {
        "CLAUDE_SONNET_API_KEY": "---redacted---",
        "CLAUDE_SONNET_MODEL": "claude-sonnet-4-10",
        "DEPRECATED_MODELS": {
            "claude-sonnet-4-6": "claude-sonnet-4-10"
        }
    }
}
```

And instead of erroring because `claude-sonnet-4-10` is a (currently) non-existant model, it succeeds and the logs indicate it's pulling 4-6.

pkulkark:
@Kelketek Good catch. Turned out that the lookup only followed one alias hop. I've fixed it now. Could you try it again and let me know if it works as expected?

ai_eval/tests/test_ai_eval.py
pkulkark: The per-render map cache lives on the XBlock instance, which the LMS/CMS should rebuild for each request, so a config change is picked up on the next request without a container restart.

Kelketek:
:+1:

- [x] I tested this: I did override swaps and they appear to work as expected.
- [x] I read through the code
- [ ] ~I checked for accessibility issues~ No user-facing issues.
- [x] Includes documentation
- [ ] ~Added to the [Code Drift](https://github.com/orgs/open-craft/projects/1) project board (for backports)~

Kelketek (approved):

PR #54, opened 2026-07-09
[BB-11007] Add meilisearch support for all 3 XBlocks

pkulkark:
Private ref: [BB-11007](https://tasks.opencraft.com/browse/BB-11007)


navinkarkera (approved):
@pkulkark Looks good

PR #57, opened 2026-07-23
feat: run action button without AI evaluation

navinkarkera:
## Description

Adds a **Run** button to the Coding XBlock that executes code without triggering AI evaluation, giving students a faster edit-run loop. The existing **Submit Code** button behaves as before (execute + AI feedback).

Also adds a notification blue-dot on the Output tab when execution results are fresh, and marks AI feedback as "(may be stale)" after a Run so students know only Submit triggers re-evaluation.

### Key Changes

- **New Run button** (visible for all languages except HTML/CSS, which updates its preview live as the user types). Clicking Run executes the code via Judge0 and updates the Output tab only — no AI evaluation is requested.
- **Output tab notification**: A blue-dot indicator appears on the Output tab when execution results are updated while the user is looking at the AI feedback tab. Clicking the Output tab clears the dot.
- **Stale-feedback indicator**: After a Run, the AI feedback tab label appends "(may be stale)" if previous AI feedback exists. The indicator disappears when Submit Code is pressed and new feedback arrives.
- **HTML/CSS shortcut**: Run is hidden for HTML/CSS (the preview already updates live on every keystroke). Stale-feedback marking still applies on Submit.
- **Bug fix**: Active-tab tracking now uses a ref (`activeTabRef`) instead of closure state, so notification logic fires correctly regardless of React batching.

## Supporting information

* `Private-ref`: https://tasks.opencraft.com/browse/BB-11067

## Testing instructions

1. Use [xblock-ai-evaluation-current.patch](https://github.com/user-attachments/files/30303672/xblock-ai-evaluation-current.patch) file and apply the patch:
   ```bash
   git apply xblock-ai-evaluation-current.patch
   ```

2. Restart tutor using `tutor dev restart`.

3. **In Studio**, add a **Coding with AI Evaluation** component to a unit. Use `mock` in Judge0 key field to make use of mock patch.

4. **In the LMS**, open the component as a student and verify:
   - A **Run** button is visible alongside Submit Code and Reset.
   - Type code and click **Run** -> output appears in the Output tab, AI feedback tab shows no new evaluation. A blue-dot notification appears on the Output tab if you switch to the AI feedback tab first.
   - After a Run, the AI feedback tab reads "AI feedback (may be stale)" if prior feedback exists.
   - Click **Submit Code** -> code is executed *and* AI feedback is requested (under the mock both return canned responses). The stale indicator disappears.
   - Switch to **HTML/CSS** language in Studio settings -> Run button is hidden; the preview iframe updates as you type in the editor. Submit Code still triggers AI feedback.
   - **Reset** clears output, feedback, and all notification states.


pkulkark (approved):
@navinkarkera LGTM 👍 

- [x] I tested this: Verified the run button works as described in the testing instructions and the requirements
- [x] I read through the code

PR #62, opened 2026-08-06
Fix field scope typo storing model and language in definition scope

SWAPI03:
## Problem

Two author-configured fields are declared with `Scope=Scope.settings` using a capital `S`:

- `model` in `AIEvalXBlock` (`ai_eval/base.py`)
- `language` in `CodingAIEvalXBlock` (`ai_eval/coding_ai_eval.py`)

`Scope=` is not the XBlock field's `scope` argument. XBlock accepts arbitrary extra keyword arguments and silently ignores unknown ones, so the intended scope never gets applied. Both fields fall back to the field default `Scope.content` (user=NONE, block=DEFINITION), while every other field in these classes uses `scope=Scope.settings` (block=USAGE).

I verified this against `XBlock==6.2.0`:

```
String(Scope=Scope.settings).scope -> Scope.content   (BlockScope.DEFINITION)
String(scope=Scope.settings).scope -> Scope.settings  (BlockScope.USAGE)
```

So the selected AI model and programming language are stored on the block **definition** instead of per **usage**. That is inconsistent with the other settings and changes how the values are serialized on OLX export/import and shared when a block is reused from a content library.

## Fix

Change `Scope=` to `scope=` on both fields. No other behavior changes.

## Tests

Added a parametrized regression test (`test_author_fields_use_settings_scope`) asserting that `model` (on all three block classes) and `language` use `Scope.settings`. Confirmed it fails on the current code (scope is `content`) and passes after the fix.

### DIFF
--- a/ai_eval/base.py
+++ b/ai_eval/base.py
@@ -47,7 +47,7 @@ class AIEvalXBlock(StudioEditableXBlockMixin, XBlock):
         values=[
             {"display_name": model, "value": model} for model in SupportedModels.list()
         ],
-        Scope=Scope.settings,
+        scope=Scope.settings,
         default=SupportedModels.GPT4O.value,
     )
 
--- a/ai_eval/coding_ai_eval.py
+++ b/ai_eval/coding_ai_eval.py
@@ -60,7 +60,7 @@ class CodingAIEvalXBlock(AIEvalXBlock):
             for language in SUPPORTED_LANGUAGE_MAP
         ],
         default=LanguageLabels.Python,
-        Scope=Scope.settings,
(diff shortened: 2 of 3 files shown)


SWAPI03:
Thanks @samuelallan72, that makes sense. I opened #69 against opencraft/ulmo. Since #60 already fixed the model field there, #69 just covers the remaining language field in CodingAIEvalXBlock (still Scope= on that branch) plus a regression test for both fields.

samuelallan72:
Thanks @SWAPI03 , I'll comment on the PR. :)

PR #65, opened 2026-08-19
[BB-11168] Fix scrolling for coaching xblock's confirmation modal

pkulkark:
## Description

This PR fixes the issue reported with the confirmation dialogue positioning (learners needed to scroll to find it in a long unit) in coaching xblock.

## Supporting information

OpenCraft Internal Jira task: [BB-11168](https://tasks.opencraft.com/browse/BB-11168)

## Testing instructions

1. Deploy this branch of the xblock
2. Create a long unit with multiple Short Answer AI blocks stacked above a Coached AI Evaluation block at the bottom of the page.
3. Click "Start again" at the bottom of the long page and verify the dialog appears centered over the coaching block itself, fully visible, right where the learner is looking.
4. The backdrop dims only the coaching block, so it no longer looks like the page froze.

PR #66, opened 2026-08-25
feat: add confirm reset to all blocks

samuelallan72:
## Description

- Rewrite the confirm reset modal for Coach AI block, using Paragon's ModalDialog
- Refactor the confirm reset modal into a shared component
- Use this modal with the other two blocks (Shortanswer AI and Coding AI)

So all three blocks should have a consistent confirm reset process, using Paragon components.

<details>

image

</details>

## Supporting information

Private-ref: https://tasks.opencraft.com/browse/BB-11118

## Testing instructions

- install this branch in a sandbox, create a course, add these blocks to advanced settings for the course
- create a component in the course for each of the three AI xblocks
- configure each with valid api keys and such, and allow reset (if applicable)
- for each xblock, enter some content and click the reset button
- confirm that clicking cancel on the modal successfully cancels without resetting
- confirm that clicking the "start over" button on the modal successfully resets everything

## Other comments

- there is a hand rolled [ConfirmDialog](https://github.com/open-craft/xblock-ai-evaluation/blob/7164e4b0f9947cefaa47ca773466941a8db4b3ad/frontend/src/coaching/CoachingStudentApp.tsx#L55) function that is now only used in a single place in the CoachingStudentApp. I recommend refactoring this to use the Paragon ModalDialog and removing the custom code. This will make things for consistent and we don't need to maintain the modal ourselves.


pkulkark:
@samuelallan72 thanks for this - having one consistent confirm flow across the three blocks is a nice cleanup. One issue I found while testing on my Ulmo devstack, which I think is a regression of the previous commit on this branch. You can check #65 for the details. Because Short Answer and Coding now use the same component, they inherit the problem too (they had no modal before).

samuelallan72:
@pkulkark could you confirm the behaviour you're seeing, and the required behaviour please? I did test that there wouldn't be an issue with the modal appearing out of the viewport here, which seemed to be the main issue before. It doesn't appear directly over the block, but it does scroll the page as expected.

If possible it would be nice to keep the vanilla behaviour of the upstream paragon components. What do you think? :)

samuelallan72:
@pkulkark ah interesting, I re-tested. It works fine for me on Firefox, but not on Chromium. I opened https://github.com/openedx/paragon/issues/4430 to report this.

I'll do a bit of investigation now to see what's going on. :)

samuelallan72:
@pkulkark I didn't get far with investigating... I did see a warning about multiple renderers using the same context provider when I click the reset button, although I'm not sure if that's related:

```
Warning: Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.
    at IntlProvider ...
    at SharedIntlProvider ...
```

frontend/src/shared/ConfirmResetModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+      inset: 0,
+      backgroundColor: 'rgba(0, 0, 0, 0.2)',
+      zIndex: 999,
+    }}>
+      <ModalPopup
+      title={title}
+      hasArrow
+      positionRef={positionRef}
+      isOpen={isOpen}
+      variant='danger'
```
pkulkark: I don't think this is getting applied. Perhaps it should be applied to the parent `Button` element?

frontend/src/shared/ConfirmResetModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+      // XXX: I couldn't use the built-in onClose functionality, because it's too sensitive.
+      // The onClose was triggered even when clicking the confirm button.
+      onClose={() => {}}
+      placement='right'
+      >
+        <div role='dialog' aria-live='polite' className="bg-white p-3 rounded shadow border">
```
pkulkark: There's no accessible name, so a screen reader announces just "dialog". Could we give the h3 an id and point `aria-labelledby` at it, and add `aria-modal="true"`?

`aria-live='polite'` shouldn't be on a dialog. It isn't a live region, and combined with the focus trap it will get announced twice.

pkulkark (approved):
@samuelallan72 Just a couple of minor nits, otherwise LGTM 👍 

- [x] I tested this: verified the modal popup works as described in the testing instructions
- [x] I read through the code
- [x] I tested for accessibility

frontend/src/shared/ConfirmResetModal.tsx
samuelallan72: Ah yes, this is leftover from when I was experimenting with ModalDialog. Nice catch. :)

frontend/src/shared/ConfirmResetModal.tsx
samuelallan72: @pkulkark thanks, this makes sense. I'll make these changes. :)

frontend/src/shared/ConfirmResetModal.tsx
samuelallan72: Added the danger variant to the confirm button:

image

samuelallan72:
@pkulkark thanks! I've addressed your comments, and will merge once tests pass. :)

PR #67, opened 2026-08-25
fix: improve error messages

samuelallan72:
## Description

Improve the generic or technical error messages when there is a failure
submitting, resetting, etc.

Also use the Paragon ModalDialog element instead of the window.alert.

<details>

image

</details>

## Supporting information

Private-ref: https://tasks.opencraft.com/browse/BB-11118

## Testing instructions

- install this branch in a sandbox, create a course, add these blocks to advanced settings for the course
- create a component in the course for each of the three AI xblocks
- configure each with valid api keys and such, and enable 'allow reset' where applicable
- for each xblock, trigger errors. I found the easiest way was to switch the tab to offline mode after loading the page (browser tools -> network -> throttling), then click the various reset, run, submit, send, etc. buttons
- verify that an accessible error message modal is displayed when an error occurs
- verify that the modal successfully is closed when clicking the close button


pkulkark (changes requested):
@samuelallan72 Could you add some tests for these error messages?

ai_eval/static/css/coding_ai_eval.css
```diff
@@ -20,7 +19,7 @@
 
 .eval-ai-container {
   display: flex;
-  max-height: 600px;
+  max-height: 610px;
```
samuelallan72: I had to do this to fix the vertical alignment of the reset/run/submit buttons in coding

ai_eval/static/css/coding_ai_eval.css
```diff
@@ -1,5 +1,4 @@
 .used-prog-language {
-  flex: 5;
```
samuelallan72: removed because it did nothing

frontend/src/coding/CodingStudentApp.tsx
```diff
@@ -429,6 +430,20 @@ export default function CodingStudentApp({
...
+      title,
+      message: message.toString(),
+    });
+    // Set a generic status message for the status area.
+    setStatusMessage(
+      intl.formatMessage({
+        id: "shortanswer.student.errorStatus",
+        defaultMessage: "An error occurred",
+      }),
+    );
```
samuelallan72: Not sure if we still want these generic status message updates when we're already providing a separate alert. I guess this will sound strange in a screenreader because both will be announced.

pkulkark:
@samuelallan72 The CI tests seem to be failing. Could you fix those?

frontend/src/coding/CodingStudentApp.tsx
pkulkark: That's a good point. I'd go with the alert as the single announcement.

frontend/src/coding/CodingStudentApp.tsx
```diff
@@ -702,16 +717,15 @@ export default function CodingStudentApp({
       } catch (error) {
         // ignore focus errors
       }
-    } catch (error: unknown) {
-      const message = getErrorMessage(
-        error,
+      setErrorData(null);  // clear the error data because this action was successful
```
pkulkark: We should do the same for short answer and coaching errors too.

pkulkark:
Nit: Could you make the dismiss button a bit more noticeable/different from the error message text?

samuelallan72:
@pkulkark 

> Nit: Could you make the dismiss button a bit more noticeable/different from the error message text?

This is the [default dismiss button in paragon](https://paragon-openedx.netlify.app/components/alert/?themes=N4IgLiBcDaoHZRAGQPYGMCGAbABAFQAsBTAWyJABoQBXKaEAsMAB0gHo2t1sCUBnMJADMABjFs0KAE5EAdGj59KDJqw5dMWXgOFiRnAJYBzRvMUgAugF8LVA1ACMVoA#variants), but I do see what you mean. Let's see what Cassie thinks. :)

samuelallan72:
@pkulkark 

> The CI tests seem to be failing. Could you fix those?

Done! The commitlint is still failing, but I'll fix that when I squash commits for merge.

I'll also add more tests when we get final confirmation of the behaviour (I think it's mainly the dismiss button now).

frontend/src/shared/error.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <Alert
+      show={true}
+      variant='danger'
+      icon={Error}
+      onClose={onClose}
+      dismissible
+      className="mt-3"
+      aria-live="polite"
```
pkulkark: We can remove `polite` here, now that there we have a single announcement.

pkulkark (approved):
@samuelallan72 Just the one comment, otherwise LGTM 👍 

- [x] I tested this: Verified the error message alerts behave as described in the testing instructions.
- [x] I read through the code.
- [x] I checked for accessibility.

frontend/src/shared/error.tsx
samuelallan72: @pkulkark I can't test this easily, but I'm wondering if we still needs this - otherwise if it alerts again, it might not get re-announced? :thinking:

